Issue #189, opened 2015-02-17
Insufficient error output during installation when output too long


donarus:
Hi,
  I think there is a problem with long exception outputs when something goes wrong during 3rd party package installation

If message contains more than `getOption("warning.length")` chars then warning/message/stop functions cuts off the end of the message and it is possible to set 8170 as a max.value for warning.length. So if the message is longer, you very often lost the most interesting part of it.

I know that there couls be a reason to keep messages short but I don't see the reason here. So I suggest to use `print` to print out the output message first and then print shortened error message.

see.: https://github.com/rstudio/packrat/blob/master/R/install.R#L569-L580

I'm working on a fix, so I will add pull request soon.

donarus:
This is already out-of-dated.. PR has been already merged and issue has been fixed.

PR #191, opened 2015-02-17
print error message during package installation into stderr in case message is too long (#189)

donarus:
fixing #189


kevinushey:
Thanks!

PR #214, opened 2015-05-11
Do not create an empty ignore file.

### LINKED ISSUE #206: packrat creates an empty .gitignore even if vcs.ignore.* are false
**Problem:** If I set both `vcs.ignore.lib` and `vcs.ignore.src` equal to `FALSE`, then I would expect packrat *not* to create an empty `.gitignore` file if there was no such file. I am managing my `.gitignore` file in a parent folder, and packrat creating its own empty `.gitignore` file overrides the one in the parent folder.

**Solution:** The fix is simply not to create an empty `.gitignore` in `updateIgnoreFile` unless there is something to add to the file.

aronatkins:
Do not create an ignore file if it does not already exist and we have no content.

Fixes #206.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -220,10 +220,12 @@ updateIgnoreFile <- function(project = NULL, file, add = NULL, remove = NULL) {
 
   project <- getProjectDir(project)
 
-  ## If the file doesn't exist, create and fill it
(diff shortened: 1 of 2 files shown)


kevinushey:
Thanks!

Issue #202, opened 2015-03-18
improve the hash function


kevinushey:
From @jcheng5:

- The GitHub SHA is used verbatim as the hash, if it's present--doesn't that omit the LinkingTo info?
https://github.com/rstudio/packrat/blob/473e73e66c283e9272beb15f48a8c6b9bfe2db29/R/cache.R#L23-L24
- I suspect the "fields" variable doesn't include enough fields--does it distinguish between GitHub and CRAN sourced packages, for example?
- Rather than recursively calling hash() on the LinkingTo packages, shouldn't we rather pull the Hash field out of their DESCRIPTION files? (assuming we always install this field into the DESCRIPTION file after we install)

PR #215, opened 2015-05-11
Consider the GitHub SHA as just another hash input.

aronatkins:
Fixes the first and second parts of #202.

The GitHub SHA is never taken verbatim as the hash but included in the content we use to compute the hash for an installed package. This also lets us distinguish between CRAN and GitHub installed packages due to the presence of `GithubSHA1` (perhaps there are more attributes?).


kevinushey:
Thanks -- I'll merge this for now but we should definitely re-visit hashing later and be sure that we're happy with how we hash.

PR #219, opened 2015-05-19
record a manifest (collection of DESCRIPTION files) with packrat projects

kevinushey:
This PR ensures that `packrat::snapshot()` calls will also record the `DESCRIPTION` file of all packages discovered and used at the time of snapshot.

### DIFF
--- a/R/manifest.R
+++ b/R/manifest.R
@@ -4,7 +4,7 @@ readManifest <- function(project = NULL) {
     return(list())
 
   content <- read(path)
-  splat <- strsplit(content, "\n{2,}", perl = TRUE)
+  splat <- strsplit(content, "\n{2,}", perl = TRUE)[[1]]
   lapply(splat, readManifestEntry)
 }
 
--- a/R/manifest.R
+++ b/R/manifest.R
@@ -1,0 +1,38 @@
+readManifest <- function(project = NULL) {
+  path <- manifestFilePath(project)
+  if (!file.exists(path))
+    return(list())
+
+  content <- read(path)
+  splat <- strsplit(content, "\n{2,}", perl = TRUE)
+  lapply(splat, readManifestEntry)
+}
+
+# Manifest entries are just DESCRIPTION files
+readManifestEntry <- function(entry) {
+  read.dcf(textConnection(entry))
+}
+
+writeManifest <- function(manifestPath, pkgRecords) {
+
+  ip <- installed.packages()
+  missingPkgs <- c()
+
+  content <- vapply(pkgRecords, FUN.VALUE = character(1), USE.NAMES = FALSE, function(record) {
+    name <- record$name
+
+    if (!(name %in% rownames(ip))) {
+      missingPkgs <- c(missingPkgs, name)
+      next
+    }
+
+    libPath <- ip[name, "LibPath"]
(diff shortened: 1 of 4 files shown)


jcheng5:
Should we move ahead with this? I'm not in love with it but I don't have a better suggestion.

kevinushey:
I don't love it either... do you think it will be suitable in the interim (at least for connect's purposes)?

I think down the road we should definitely have the lockfile just record the entirety of package DESCRIPTIONs, but that might take a while to implement + test + do in a backwards compatible way.

jcheng5:
Yes, let's go with it--if somehow @aronatkins isn't able to solve the hashing issues with it (which seems like an unlikely scenario) we can remove it before the next CRAN release.

aronatkins:
Having a manifest with full `DESCRIPTION`s is just the first piece, right? `packrat::restore` needs to take advantage of that content to compute package hashes on-the-fly instead of using the hashes in `packrat.lock`.

@jcheng5 Is your current thinking that should be creating a `packrat.lock` on the server before calling `packrat.restore`? `packrat:::hash` is not currently appropriate for this situation, as it uses `system.file` to discover `DESCRIPTION` files for all recursive references. I guess I'm a little unclear about what is supposed to happen within vs. outside `packrat::restore`.

jcheng5:
@aronatkins For some reason I had it in my head that you were going to calculate hashes yourself (outside of packrat), now that I think about it I don't know why that's a good idea though.

aronatkins:
@jcheng5 Glad to know that we both had the same, "why does this make sense?" moment. :ghost:

jcheng5:
@aronatkins I seem to have a lot of those lately :-\

jjallaire:
Just to clarify -- RStudio Connect has it's own copy of packrat, correct?
Just making sure because it's very difficult for us to ensure that the most
recent version of packrat (especially one on GitHub) is sent up by the IDE.

jcheng5:
Talking with @jmcphers and @kevinushey we came up with the conclusion that this manifest data should be transmitted basically in the form that this PR records it, but rather than having it on disk, we should generate it on the fly during deployment, and take it as an optional piece of data during restore. This avoids having a second source of truth being on disk/in source control, where it could potentially get out of sync.

The only "downside" to having this information be ephemeral instead of persistent is that the "other" use of packrat besides deployment--making projects reproducible across time and space--won't be able to take advantage of this data. We decided that wasn't so important, because in that scenario we could use .Rprofile to enforce that the same version of Packrat be used no matter where the project is used. (Actually come to think of it, it's possible that a `git pull` followed immediately by a packrat command would cause problems--but at least there's a way out, that's as simple as restarting R.)

@kevinushey and I will work together on this either later this week or early next week, we will also keep @aronatkins abreast & involved.

aronatkins:
@jcheng5 and I just chatted; we may be able to take advantage of the `manifest.json` as input.  We can totally parse the `manifest.json` on the server-side and then pass the `manifest$packages` sub-document to `packrat::restore`.

Nice side-effect/next-step: packrat wouldn't need to download the package source before computing hash for hit/miss. We could avoid source download completely because the `manifest.json` provides complete DESCRIPTION data.

jcheng5:
Obsoleted by a different approach, see #225.

PR #225, opened 2015-06-11
Allow hashes to be calculated based on non-installed DESCRIPTION files

jcheng5:
Related:
https://github.com/rstudio/connect/pull/514 (see here for summary)
https://github.com/rstudio/rsconnect/pull/42

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -13,11 +13,7 @@ installedDescLookup <- function(pkgName) {
 
 # We assume 'path' is the path to a DESCRIPTION file
 #' @importFrom tools md5sum
-hash <- function(path, descLookup = NULL) {
-
-  if (is.null(descLookup)) {
-    descLookup <- installedDescLookup
-  }
+hash <- function(path, descLookup = installedDescLookup) {
 
   if (!file.exists(path))
     stop("No DESCRIPTION file at path '", path, "'!")
--- a/R/cache.R
+++ b/R/cache.R
@@ -7,17 +7,23 @@ isUsingCache <- function(project) {
   isTRUE(get_opts("use.cache", project = project))
 }
 
+installedDescLookup <- function(pkgName) {
+  system.file("DESCRIPTION", package = pkgName)
+}
+
 # We assume 'path' is the path to a DESCRIPTION file
 #' @importFrom tools md5sum
-hash <- function(path) {
+hash <- function(path, descLookup = NULL) {
 
+  if (is.null(descLookup)) {
+    descLookup <- installedDescLookup
+  }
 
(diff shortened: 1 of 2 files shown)


jcheng5:
If we want to call `.hash` from rsconnect, maybe it should be renamed to `.hash` and exported?

aronatkins:
Bump the package version. :+1: otherwise.

I'm OK with renaming `packrat:::hash` to `packrat::hash` but don't see why you prefer `.hash`.

aronatkins:
After mulling this a bit...

Maybe `hash` should be based off the `DESCRIPTION` data-frame rather than a path to a `DESCRIPTION` file; then we get the ability to change how we obtain that description data without further changing packrat (and pushing new versions to CRAN).

kevinushey:
Why does `hash` need to be exported? Can't you just call it with `:::`? (Or is this to avoid R CMD check grumpiness?)

R/cache.R
```diff
@@ -7,17 +7,23 @@ isUsingCache <- function(project) {
...
 }
 
+installedDescLookup <- function(pkgName) {
+  system.file("DESCRIPTION", package = pkgName)
+}
+
 # We assume 'path' is the path to a DESCRIPTION file
 #' @importFrom tools md5sum
-hash <- function(path) {
+hash <- function(path, descLookup = NULL) {
```
kevinushey: Why not just use `descLookup = installedDescLookup`?

kevinushey:
FWIW, if you want to 'sneakily' use a non-exported function and dodge R CMD check, you can probably use something like:

```R
extract_from <- function(package, object) {
    do.call(":::", list(as.name(package), as.name(object)))
}

hash <- extract_from("packrat", "hash")
```

Although maybe just exporting `.hash` has a (hidden) export would be fine.

jcheng5:
I'm fine with hash not being exported. (Not going to CRAN with the calling code) We have a ::.snapshotImpl that's called from the same place in connect, so I was wondering if I should follow that pattern.

kevinushey:
My nits early are minor enough that I'm happy merging as-is. Anything else you wanted to change?

jcheng5:
Yeah, one more round tonight to incorporate Aron's feedback.

jcheng5:
@kevinushey @aronatkins Ready for review again.

@kevinushey Do you mind bumping the version right after the merge, so @aronatkins can update Connect's minimum packrat version requirement?

aronatkins:
:mouse: I'm good.

jcheng5:
@kevinushey Never mind, I'm on it

Issue #224, opened 2015-06-10
snapshot() very slow


krlmlr:
In a project with 80 packages, `snapshot()` takes roughly two minutes on my machine. I think the reason is the recursive call to `getPackageRecords()` which seems to walk *all* dependency trees for *each* dependency. This means enumeration of all possible paths, and is troublesome e.g. for diamond structures:

A -> B -> D
A -> C -> D
D -> E -> G
D -> F -> G
G -> H -> J
G -> I -> J
...

Perhaps a caching environment can help: this env holds all package records that have been computed so far, should be consulted before doing any computations, and populated with already computed package records. However, the vectorized nature of `getPackageRecords` complicates things, a mapping from `pkgNames` to `allRecords` will be required.

kevinushey:
I agree this is a good idea. This is an operation where vectorization really doesn't give us huge benefits, and so I agree just using a for loop + a caching environment would be a better solution.

krlmlr:
Thanks. I'll look into it.

IrfyNaz:
+1 - also seeing very slow snapshots on a project with ~ 100 packages, takes 20 mins to snapshot.

PR #226, opened 2015-06-11
use names in getPackageRecords, avoid returning duplicates

krlmlr:
This is a first step towards #224, I hope this will allow implementing the cache while keeping the vectorized nature of getPackageRecords. It works by giving names to `pkgNames` which are then propagated to `allRecords`. As a bonus, `allRecords` is reordered in the order of `pkgNames`. If this introduces errors, they are not caught by the tests.

If the naming fails to carry through, would a warning/error be appropriate here?

What is a good place for `vecdiff`?


kevinushey:
I think it's okay to leave `vecdiff` there, although it could also fit in `utils.R` (either is fine to me).

That said, I don't quite understand directly from reading the code -- what is the gain of changing from `setdiff` to `vecdiff`?

krlmlr:
I can insert a second `dropNull` if this is a valid condition. Is it? I thought that it's [an error](https://github.com/rstudio/packrat/pull/226/files#diff-248d140618b0a0963077f113fa13f3b8L214) if there's no package record to a queried package, but then I don't know the implementation well enough.

We might rather want to look for empty names in `allRecords` -- shall we warn or fail in this case? (A failure in such a case indicates an implementation error.)

`vecdiff` keeps names, `setdiff` doesn't (due to the call to `unique`). Will move to `utils.R`.

krlmlr:
I have moved `vecdiff()` to `utils.R`, but I'm still unsure what to do if the names in `allRecords` for some reason don't refer to valid package names, or if no package record has been found for a package. (The former might indicate an implementation error, the latter might be a regular error condition or could be simply ignored.)

krlmlr:
Merged with master.

kevinushey:
@krlmlr sorry it's taken so long to review + get back to this. This change looks good to me; I'll merge now.

PR #230, opened 2015-07-07
use secure download method for interactions with CRAN when required

jjallaire:
Two changes here:

1) Update the `download` method to take advantage of the new secure R 3.2 download methods ("wininet" and "libcurl") when available.

2) Add a new `secureDownloadMethod` function which sets the default R download method to a secure one (if possible) when one of the repositories in `getOption("repos")` uses https. This allows `available.packages` and `install.packages` to work correctly in the presence of https repository URLs.

@jcheng5 @trestletech @wch

Winston, I'll put in a similar PR into downloader soon (packrat includes a copy of the the downloader::download function so as not to take a dependency on any other packages).

### DIFF
--- a/R/downloader.R
+++ b/R/downloader.R
@@ -37,32 +37,49 @@ download <- function(url, ...) {
   # First, check protocol. If http or https, check platform:
   if (grepl('^https?://', url)) {
 
-    # If Windows, call setInternet2, then use download.file with defaults.
+    # Check whether we are running R 3.2 and whether we have libcurl
+    isR32 <- getRversion() >= "3.2"
+
+    # Windows
     if (.Platform$OS.type == "windows") {
 
-      # If we directly use setInternet2, R CMD CHECK gives a Note on Mac/Linux
-      seti2 <- `::`(utils, 'setInternet2')
+      if (isR32) {
+        method <- "wininet"
+      } else {
 
-      # Store initial settings, and restore on exit
-      internet2_start <- seti2(NA)
-      on.exit(suppressWarnings(seti2(internet2_start)))
+        # If we directly use setInternet2, R CMD CHECK gives a Note on Mac/Linux
+        seti2 <- `::`(utils, 'setInternet2')
 
-      # Needed for https. Will get warning if setInternet2(FALSE) already run
-      # and internet routines are used. But the warnings don't seem to matter.
-      suppressWarnings(seti2(TRUE))
+        # Check whether we are already using internet2 for internal
+        internet2_start <- seti2(NA)
+
+        # If not then temporarily set it
+        if (!internet2_start) {
+          # Store initial settings, and restore on exit
+          on.exit(suppressWarnings(seti2(internet2_start)))
+
+          # Needed for https. Will get warning if setInternet2(FALSE) already run
+          # and internet routines are used. But the warnings don't seem to matter.
+          suppressWarnings(seti2(TRUE))
+        }
+
+        method <- "internal"
+      }
 
       # download.file will complain about file size with something like:
       #       Warning message:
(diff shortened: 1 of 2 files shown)


R/packrat-mode.R
```diff
@@ -182,6 +182,10 @@ afterPackratModeOn <- function(project,
     options(repos = repos)
   }
 
+  # Set a secure download method if any of the repos URLs use https
```
jcheng5: Is this something we should do every time, or only if we actually set the repos in line 182?

R/downloader.R
```diff
@@ -37,32 +37,49 @@ download <- function(url, ...) {
...
-      suppressWarnings(download.file(url, ...))
+      suppressWarnings(download.file(url, method = method, ...))
 
     } else {
-      # If non-Windows, check for curl/wget/lynx, then call download.file with
+      # If non-Windows, check for libcurl/curl/wget/lynx, then call download.file with
       # appropriate method.
 
-      if (nzchar(Sys.which("wget")[1])) {
+      if (isR32 && capabilities("libcurl")) {
```
kevinushey: `capabilities` will return a length-zero logical vector if there is no such capability called `libcurl`, so I think we need a wrapper function to ensure we only get `TRUE` / `FALSE` results.

jjallaire: I check for R 3.2 first in the conditional which I believe ensures that
capabilities("libcurl") will be there. Is that correct?

kevinushey: I guess since the check only occurs after an `isR32` call it's probably fine?

jjallaire: I should add that I've tested this code under R 3.2 and previous versions
on all platforms and there were no issues with capabilities("libcurl")

jjallaire: Yes, that's correct.

jcheng5:
Is there a reason that the downloader code doesn't just use secureDownloadMethod as well? It's basically the same logic, right?

jjallaire:
Let me take a crack at refactoring to share the logic (there are some
differences but I think we can make them orthogonal to secureDownloadMethod)

jjallaire:
It's arguably more complicated to refactor as the downloader method has a
bunch of extra stuff it does:

1. Actually forcing setInternet2 (just for the stack frame) if it's not set

2. Suppressing warnings on Windows

3. Temporarily overriding the download.extra.options to cause curl to
follow redirects

The resulting code is actually harder to understand because of the special
casing (i.e. for Windows, don't quite believe secureDownloadMethod when it
says it can't get a secure method, rather go ahead and force setInternet2,
etc.).

Part of this results from the fact that they are different use cases (i.e.
wrap download.file and do whatever you need to do for secure downloads vs.
create a global side effect that can be relied on by install.packages and
available.packages).

I think the refactor also carries more defect and regression risk as it's
brand new code.

jcheng5:
OK, LGTM then.

R/packrat-mode.R
jjallaire: I think the fix is still required (if we don't set the repos however it still has https URLs then we need to set the download method anyway)

jjallaire:
@kevinushey Realize that you haven't commented definitively I am going to merge this now because RStudio Connect requires this code to function correctly in the face of a client that has provided an HTTPS URL.

aronatkins:
Should this have bumped the version?

jcheng5:
Yes, needs a version bump, and connect should take a dependency on that new
version. Good catch.

jjallaire:
I just bumped the version.

Last CRAN submission was just over two weeks ago. Should we push to have
another release immediately or just wait for the ordinary cycle?

I'd probably lean towards waiting as there is an easy workaround (just
configure your server with a download method that supports https).

aronatkins:
I think we're OK holding for now.

PR #235, opened 2015-07-29
Allow YAML headers to end with dots

noamross:
See https://github.com/rstudio/rmarkdown/pull/485

All tests pass


jjallaire:
@kevinushey This looks good to me. Could you do another sanity check and merge?

kevinushey:
Thanks!

PR #239, opened 2015-08-28
ensure downloader respects HTTP error codes

kevinushey:
NOTE: This PR will need some pretty thoughtful review to ensure it's correct...

There are two overarching issues that this PR attempts to resolve:

1) When an external program (e.g. `wget`, `curl`) fails to download a file, the status code from running the program is logged as a warning when non-zero, but no error is emitted. This is in contrast to the behaviour of `method = "internal"`, which _does_ err in such a case.

2) With R (SVN revision < 69197), `libcurl` also does not respect HTTP error codes; even worse, it downloads the 404 page and returns a 0 (success) status code. This is only fixed in recent R SVN revisions.

For 1), we catch warnings from `download.file()` when using methods `curl` and `wget`, and err if a non-zero status code is returned.

For 2), we only enable `libcurl` on versions of R with SVN revision >= 69197.

In the end, this fixes a very ugly problem where `snapshot()` could silently download a 404 page and pretend that the 404 page was the source package it intended to download.

### DIFF
--- a/R/downloader.R
+++ b/R/downloader.R
@@ -33,9 +33,7 @@
 #          "downloader.zip", mode = "wb")
 # }
 #
-download <- function(url, ...) {
-
-  method <- inferAppropriateDownloadMethod(url)
+download <- function(url, method = inferAppropriateDownloadMethod(url), ...) {
 
   # When on Windows using an 'internal' method, we need to call
   # 'setInternet2' to set some appropriate state.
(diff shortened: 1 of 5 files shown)


kevinushey:
@wch, I think some of these ideas may want to be incorporated into your [downloader](https://github.com/wch/downloader) package?

wch:
That sounds like a good idea.

PR #250, opened 2015-10-14
Add implicit.packrat.dependency parameter to snapshotImpl

kippandrew:
By default packrat itself is included as an implicit dependency by `snapshotImpl`, there are use cases (`rsconnect`) where this is not desirable. This PR adds a parameter to disable this dependency by passing `implicit.packrat.dependency = FALSE` to `snapshotImpl`

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -11,6 +11,7 @@
 #'   directory.
 #' @return Returns a list of the names of the packages on which R code in the
 #'   application depends.
+#' @param implicit.packrat.dependency is packrat an implicit dependency?
 #' @details Dependencies are determined by parsing application source code and
 #'   looking for calls to \code{library}, \code{require}, \code{::}, and
 #'   \code{:::}.
@@ -29,7 +30,8 @@
 #' @keywords internal
 appDependencies <- function(project = NULL,
                             available.packages = NULL,
-                            fields = c("Imports", "Depends", "LinkingTo")) {
+                            fields = c("Imports", "Depends", "LinkingTo"),
+                            implicit.packrat.dependency = TRUE) {
 
   if (is.null(available.packages)) available.packages <- available.packages()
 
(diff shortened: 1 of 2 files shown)


jjallaire:
@kippandrew If your rsconnect PR requires this PR then we'll need to do some orchestration before we merge the rsconnect PR (this is because RStudio IDE currently embeds rsconnect master so once this binding exists we'd also need to embed packrat master which is I think a bit too aggressive). We're due for a packrat release to CRAN soon anyway, so should be able to sort it out soon.

aronatkins:
Would be nice to see a test prove packrat is excluded; maybe a variant of https://github.com/rstudio/packrat/blob/f6bf32140fb8a8e9686bb62b791a3b1df2e2f0dc/tests/testthat/test-packrat.R#L61-L84

kippandrew:
@jjallaire The rstudio/rsconnect#76 PR doesn't need this, but I did want to bump the minimum required version of `packrat` for `rsconnect` because there was a bug with Bioconductor packages (#247) that was recently fixed. I hit it in my testing so I imagine others will be bitten by it.

jjallaire:
Okay, so let's get packrat onto CRAN next week (Kevin is out until Tuesday)
then we can merge the rsconnect PR. BTW that diff is hard to reason about
(probably because a bunch of code was removed not because it's particularly
complicated). I'd suggest a walkthrough with Jonathan as a sanity check (as
once we merge it it's going "hot" into the preview release which will be
widely promoted next week).

kippandrew:
@jjallaire I'm fine holding off on that PR until after the preview release drops. There are server sides changes that need to go out before the `rsconnect` changes can land.

kevinushey:
This PR looks good to me; should I merge or hold off for orchestration of other things?

jjallaire:
This is fine to merge (other things depend on this PR rather than the other
way around)

PR #268, opened 2016-02-11
Bugfix in getSourceForPkgRecord.

CedricFch:
Try downloading archive versions when package is not available (older versions of package could be available).

Details:
When the current version of a package depends on some version of R, it is not returned by the function available.packages if you are running an older R version. However, an archive version of the package could be just fine. This is why I suggest that we should try downloading archive versions of unavailable packages.
Any thoughts about this ?

### DIFF
--- a/R/restore.R
+++ b/R/restore.R
@@ -112,12 +112,13 @@ getSourceForPkgRecord <- function(pkgRecord,
       }
     })
     type <- "local"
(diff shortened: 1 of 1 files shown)


JeanLescut:
Good work !

kevinushey:
Thanks @CedricFch -- sorry for the late response! I'm going to make a similar commit on master with this same fix plus some other tweaks.

PR #271, opened 2016-02-19
use devtools downloader when available

kevinushey:
This PR modifies the packrat downloader to use `devtools::download()` when available.


kevinushey:
Hmm, `devtools:::download()` is just a wrapper around some `httr` calls; we likely need to just write our own functions that respect `devtools:::github_pat()` or something to that effect.

PR #280, opened 2016-03-04
detect Rmd dependencies based on custom output formats

jjallaire:
Currently if a custom output format from another package is used in an Rmd it is not detected as a dependency, e.g. for

```yaml
output: mypackage::myformat
```

`mypackage` is not detected as a dependency.

This PR calls into the `rmarkdown::default_output_format` function to determine if there is a custom format implemented within a package and adds that package to the dependencies.

Note that there is already some processing of the yaml within Rmd files however this is all based regular expressions and determining the default format would involve both much more complex parsing (using the yaml package) as well as doing a bunch of other resolution (e.g. looking for an `_output.yml` file in the Rmd file's directory) so it's much better to call into the rmarkdown package for this.

Note that 3 tests failed for me locally after applying this change but I'm not sure if I'm well equipped to diagnose whether this change caused the failures or not.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,7 +17,8 @@ Suggests:
     testthat (>= 0.7),
     devtools,
     httr,
-    knitr
+    knitr,
+    rmarkdown
 Enhances: BiocInstaller
 URL: https://github.com/rstudio/packrat/
 BugReports: https://github.com/rstudio/packrat/issues
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -160,6 +160,15 @@ fileDependencies.Rmd <- function(file) {
 
   deps <- "rmarkdown"
 
+  # check whether the default output format references a package
+  if (requireNamespace("rmarkdown", quietly = TRUE)) {
+    format <- rmarkdown::default_output_format(file)
+    components <- strsplit(format$name, "::")[[1]]
+    if (length(components) == 2) {
+      deps <- c(deps, components[[1]])
+    }
+  }
+
   # We need to check for and parse YAML frontmatter if necessary
   yamlDeps <- NULL
   content <- readLines(file)


kevinushey:
@jjallaire I'm going to merge this and make the necessary fixes on master. Thanks!

kevinushey:
@jjallaire the errors were due to a relative path not being normalized. Since the `rmarkdown` package munges the working directory before doing work, should `default_output_format()` also normalize the path as well? https://github.com/rstudio/rmarkdown/blob/master/R/output_format.R#L314-L332

jjallaire:
Sorry I don't fully understand the question. rmarkdown just needs to
determine the directory of the input file -- it seems like it's able to do
so at first glance. What potential issue are you seeing?

kevinushey:
Here's what I mean:

```R
> setwd("tests/testthat")
> getwd()
[1] "/Users/kevinushey/git/packrat/tests/testthat"
> rmarkdown::default_output_format("resources/params-example.Rmd")
Error in file(con, "r") : cannot open the connection
In addition: Warning message:
In file(con, "r") :
  cannot open file 'resources/params-example.Rmd': No such file or directory
> setwd("resources")
> rmarkdown::default_output_format("params-example.Rmd")
$name
[1] "html_document"
...
```

jjallaire:
I think I'm being dense but I still don't fully understand. What would be
the proposed change?

In general since rmarkdown just went to CRAN it would be great if this
didn't require a change to work correctly (and not produce an error). I
guess worst case if we have to make a change in rmarkdown we can version
check before calling. Again though I'd like to avoid this as we're going to
be promoting a new dashboard format (http://rstudio.github.io/flexdashboard/)
which won't deploy correctly to shinyapps.io or Connect without either this
change OR we just tell everyone to do library(flexdashboard) in their setup
chunk (not the end of the world).

kevinushey:
The proposed change would have `rmarkdown::default_output_format()` handle relative paths, e.g. `"foo/bar.Rmd"`, but callers of that function can easily work around the error by either providing a normalized path or moving to directory `foo` on their own.

In short it's nothing urgent for `rmarkdown`, client packages just need to be a bit careful in how they call `rmarkdown::default_output_format()` at the moment.

jjallaire:
Okay, so is packrat covered for now?

I wrote up an issue so we remember to handle this in rmarkdown as well:
https://github.com/rstudio/rmarkdown/issues/637

kevinushey:
Yep, all is well on the packrat side.

PR #283, opened 2016-03-14
remove install_github shim

jjallaire:
We originally had a goal of not requiring users to have devtools as part of their packrat managed dependencies (because it in turn brings in a whole bunch of other dependencies). We did this by attempting to temporarily load and call devtools. However, in my experience this was flaky and I ended up always installing devtools and it's dependencies within my project. With further dependencies on devtools for private GitHub repos I think it's much better if eliminate the shim and make the dependency much more explicit.

Note that we should still have it as a goal to break the devtools installation functions into their own package so that devtools is free to add dependencies without worrying about their footprint in packrat-managed projects.

I ran R CMD check with this PR and did see one error (but it appears unrelated):

```
  Running ‘test-all.R’ ERROR
Running the tests in ‘tests/test-all.R’ failed.
Last 13 lines of output:
                  library   packrat
      breakfast     1.0.0        NA
  
  Use packrat::snapshot() to record these packages in packrat.
  testthat results ================================================================
  OK: 92 SKIPPED: 1 FAILED: 1
  1. Error: We can create and 'upload' an example package 
  
  Error: testthat unit tests failed
  In addition: Warning message:
  In install.packages("sashimi", lib = tempLib) :
    installation of package 'sashimi' had non-zero exit status
  Execution halted
```


kevinushey:
That test failure does indeed seem unrelated (unfortunately, the repository tests are finicky as they try to run `R CMD check` from an R process, and this does not always work well when run under devtools / testthat.)

Issue #288, opened 2016-04-01
Symlinking not working on latest version


raubreywhite:
my code:

setwd("'H:/xSmallProjects/test")
packrat::init()

Everything then works, it gets down to "Initialization complete!", then restarts R and I get this error:

Error in file(con, "rb") : cannot open the connection
In addition: Warning message:
In file(con, "rb") :
  cannot open file 'H:/xSmallProjects/test/packrat/lib-R/base/DESCRIPTION': No such file or directory

I just rolled back my version of packrat to 0.4.4 and everything works perfectly.

kevinushey:
Thanks! This should be resolved with https://github.com/rstudio/packrat/commit/2bbc359d5d53f9c5061ffd48ec2b770532d693e0 (as per your comments).

PR #289, opened 2016-04-04
Fixes bug #288 - symlinking windows on networked drives

raubreywhite:
https://github.com/rstudio/packrat/issues/288

This is a windows problem.

R/RStudio is on the C drive, and my project folder is on a networked drive. Windows will not allow a symlink/junction between the two drives, so instead of having a bunch of links, such as: packrat/lib-R/base I have a bunch of empty folders named: packrat/lib-R/base. Thus when "isPathToSamePackage" is called, we try to read packrat/lib-R/base/DESCRIPTION (which does not exist) and an error happens. This results in killing packrat, instead of returning a "FALSE" like it should.

Another option would be to edit the function "isPathToSamePackage" and include a "if(!file.exists(lhsPath) | file.exists(rhsPath)){ return(FALSE) }" before the "readChar" commands.

### DIFF
--- a/R/library-support.R
+++ b/R/library-support.R
@@ -70,9 +70,13 @@ ensurePackageSymlink <- function(source, target) {
   # library corresponding to the current running
   # R session.
   if (file.exists(target)) {
-
-    if (isPathToSamePackage(source, target))
-      return(TRUE)
+    fail <- TRUE
+    try({
+      if (suppressWarnings(packrat:::isPathToSamePackage(source, target)))
+        return(TRUE)
+      fail <- FALSE
+    },TRUE)
+    if(fail) return(FALSE)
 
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks for the PR -- I think I prefer your second change, so I'll make a commit to that effect. Sorry for the trouble!

raubreywhite:
Thanks!

I just noticed one unintended consequence of the second change. In library-support.R (bbc359), line 83 needs to be "unlink(target, **recursive=TRUE**)", as windows displays the failed junctions as empty folders, and the recursive argument is needed to delete them.

kevinushey:
Thanks -- that should now be addressed with https://github.com/rstudio/packrat/commit/cc74baa959e4a6e10bef4ec1df79c9fd902cd971

PR #293, opened 2016-04-13
Checking for failed symlinks

raubreywhite:
In Windows, if the symlink fails (e.g. on a networked drive), an empty folder will be created. So by just checking if the folder exists, we aren't actually checking if the symlink was successful. We need to go inside the folder and see if the files actually exist like we expect them to.


kevinushey:
Thanks!

PR #297, opened 2016-05-13
update install instructions to CRAN

bquast:
As a sidenote, I just had a first run with profvis, the use of package::function is actually very costly in terms of time, I am not sure this so such a good practice. In this case, in the documentation of a package, I think it is safe to assume the user can load the package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,7 @@ new in Packrat.
 
 Start by installing Packrat:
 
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks!

kevinushey:
Also, although there is definitely a bit of overhead with explicitly qualifying the namespace for calls, this generally isn't enough to matter except with very computationally expensive routines.

PR #315, opened 2016-08-10
packrat.lock repos are used when testing if a package has CRAN-like source

aronatkins:


kevinushey:
LGTM, thanks! (In theory the value of `getOption("repos")` should be consistent with the lockfile as long as packrat mode is 'on'; I guess this isn't true in the Connect case since we're just directly going to `packrat::restore()`? Just want to make sure I understand the issue.

aronatkins:
@kevinushey You've got it. We directly call `packrat::restore` and do nothing more.

Basically (with some unrelated code deleted):
```r
Sys.setenv("R_PACKRAT_SRC_OVERLAY"=config$SourcePackageDir)
Sys.setenv("R_PACKRAT_CACHE_DIR" = packratCacheDir)

packrat::set_opts(
  auto.snapshot = FALSE,
  use.cache = TRUE,
  project = getwd(),
  persist = FALSE
)

packrat::restore(overwrite.dirty=TRUE, prompt=FALSE, restart=FALSE)
```

PR #316, opened 2016-08-11
set the repos option during restore so it is available downstream

aronatkins:
This allows successful, repeated `packrat::restore` when using a private repository.

Builds on #315 

The one thing I dislike about this approach is that we have a mixture of `repos` passed as a function argument and `repos` available via `getOption("repos")`. This inconsistency can be confusing when reading through the code. It would be nice to shift to one model or the other.

### DIFF
--- a/R/packrat.R
+++ b/R/packrat.R
@@ -390,9 +390,17 @@ restore <- function(project = NULL,
     pkgsToIgnore <- dirtyPackageNames[!dirtyPackageNames %in% pkgNames(packages)]
   }
 
+  # Configure repos globally to avoid explicitly passing the repos list to all
+  # downstream function calls.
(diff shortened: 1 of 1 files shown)


aronatkins:
@kevinushey I have not added tests for either this PR or #315. Do you have suggestions on how to approach testing restore with private repos?

kevinushey:
Unfortunately we don't have the helpers set up for easy testing here, but basically we'd need some code to generate a custom CRAN-like repository with some packages, and a packrat lockfile that indicates a particular package should be recovered from a CRAN-like repository.

I think we effectively test this in the case where packrat mode is 'on' (and so the `repos` option has already been suitably populated); this PR just effectively ensures that we're always on this code path, even if packrat mode is not 'on' for a particular repository.

Issue #325, opened 2016-08-22
adding packages to cache can incur cross-device copy


aronatkins:
When using the packrat package cache, packages are first installed into the `packrat/lib` directory. After that install, they are subsequently moved to the cache and a symlink created in place of the original directory.

This presents a problem when the cache is on a separate mount than the `packrat/lib` directory. Adding to the cache is not simply a `file.rename` but a full directory copy.

Connected to #324

kevinushey:
This one is unfortunately going to be tough to solve in general. It's possible that we could optimize the directory copy performance however, for the case where we just want to copy one directory from some location to another. (Right now we do a file-by-file copy, which is likely incurs some overhead)

aronatkins:
I was hoping that we would never install the package into `packrat/lib` if the cache was in-use, but instead in `CACHE_DIR/UNIQUE_DIRNAME` and then perform a rename inside `CACHE_DIR` before the final symlink. This also helps with #324 because we will fix-up the symlinks as we encounter packages rather than at the end.

In other words: If you know the package is destined for the cache, start by putting it there.

This is a larger change than just optimizing the copies.

PR #326, opened 2016-08-22
messaging about enrolling packages in the package cache.

aronatkins:
When there are lots of newly installed packages and they need to be added to
the package cache, promoting these packages to the cache can take lots of time
(due to #325). Add some messaging for transparency into this process.

Related to #325 (but does not address it)

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -94,6 +94,10 @@ normalizeForHash <- function(item) {
   gsub("[[:space:]]", "", paste(unlist(item), collapse = ""))
 }
 
+isVerboseCache <- function() {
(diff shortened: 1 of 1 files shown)


aronatkins:
I'm not in love with any of this language; feel free to adjust as you see fit. I mostly care about transparency into what work `packrat::restore` is performing.

kevinushey:
How about hiding these behind an option `packrat.verbose`, just to ensure that packrat isn't too chatty for users who don't want to see these messages?

kevinushey:
Thanks!

PR #346, opened 2016-12-09
symlink packages as they are added to the cache

### LINKED ISSUE #345: initial package install populates cache but does not symlink
When `packrat::restore` initially installs a package into the cache, the `packrat/lib` directory still contains the actual package source and is not converted into a symlink.

c35c5b39651989845a2501c80516e75ad7f247a2 does not have this problem (symlinks are created)

3288b5554d92278a50a0da738ce13bf64ba193d6 does not create symlinks.

```
packratLock = c(
  "PackratFormat: 1.4",
  "PackratVersion: 0.4.4",
  "RVersion: 3.2.0",
  "Repos: CRAN=http://cran.rstudio.com",
  "",
  "Package: jsonlite",
  "Source: CRAN",
  "Version: 0.9.16",
  "Hash: 24d362a166a96ac1969e64ede522c717"
)

# teardown from prior run
if (dir.exists("no_symlink")) {
  unlink("no_symlink", recursive = TRUE)
}

here <- getwd()

# setup
dir.create("no_symlink")
setwd("no_symlink")
test_root <- getwd()

dir.create("cache")
dir.create("app")
dir.create("app/packrat")

fp <- file("app/packrat/packrat.lock")
writeLines(packratLock, fp)
close(fp)

setwd("app")
Sys.setenv(R_PACKRAT_CACHE_DIR = file.path(test_root,"cache"))
options(packrat.verbose.cache = TRUE,
        packrat.connect.timeout = 10)
packrat::set_opts(
  auto.snapshot = FALSE,
  use.cache = TRUE,
  project = file.path(test_root,"app"),
  persist = FALSE
)

packrat::restore(overwrite.dirty = TRUE,
                 prompt = FALSE,
                 restart = FALSE)

rv <- R.Version()
package_dir <- file.path(test_root,
                        "app",
                        "packrat",
                        "lib",
                        rv$platform,
                        getRversion(),
                        "jsonlite")

if (!packrat:::is.symlink(package_dir)) {
  cat("FAIL: package directory ", package_dir, " is not a symlink\n", sep = "")
} else {
  cat("PASS: package directory ", package_dir, " is a symlink\n", sep = "")
}

setwd(here)
```

Running like:

```
source("no_symlink.R", chdir = TRUE)
```

Produces output like:

```
Installing jsonlite (0.9.16) ... 
	OK (built source)
Promoting newly installed packages to the package cache.
Caching jsonlite.
FAIL: package directory /Users/aron/dev/rstudio/packrat/no_symlink/app/packrat/lib/x86_64-apple-darwin13.4.0/3.2.4/jsonlite is not a symlink
Warning message:
In packrat::restore(overwrite.dirty = TRUE, prompt = FALSE, restart = FALSE) :
  The most recent snapshot was generated using R version 3.2.0
```

aronatkins:
now that we no longer have a separate pass to fix all the symlinks, we need to
take care of that as we add packages into the cache.

Fixes #345


aronatkins:
`moveInstalledPackagesToCache` is obsolete as of 6b27f2511fd866b0cb2556a2428797dec0b6b0dd; not removing it here.

R/cache.R
```diff
@@ -116,14 +123,23 @@ moveInstalledPackageToCache <- function(packagePath,
   cachedPackagePath <- cacheLibDir(packageName, hash, packageName)
   backupPackagePath <- tempfile(tmpdir = dirname(cachedPackagePath))
 
+  # helper function to remove the package from its original location and
+  # create a symlink to the cached version.
+  packageSymlink <- function() {
+      unlink(packagePath, recursive = TRUE)
+      symlink(cachedPackagePath, packagePath)
```
kevinushey: Can we handle a failed symlink here? Right now if the symlink fails then the application will be busted (the package will just be gone from the private library).

I would prefer something like:

1. Move package installation to temporary path in same directory,
2. Attempt a symlink,
3. On success, remove the directory produced in 1),
4. On failure, move the directory produced in 1) back home.

tests/testthat/test-cache.R
```diff
@@ -0,0 +1,55 @@
+context("Cache")
+
+withTestContext({
+  # https://github.com/rstudio/packrat/issues/345
+  test_that("package installation when configured with a a cache uses the cache", {
+    skip_on_cran()
```
kevinushey: We should also skip on Windows here.

tests/testthat/test-cache.R
```diff
@@ -0,0 +1,55 @@
...
+    # non-symlink. dir.exists only tells us if the package is installed, as it
+    # reads through symlinks.
+    expect_true(dir.exists(package_dir),package_dir)
+    expect_false(packrat:::is.symlink(package_dir), package_dir)
+
+    Sys.setenv(R_PACKRAT_CACHE_DIR = theCache)
+    on.exit(Sys.unsetenv("R_PACKRAT_CACHE_DIR"))
+
+    packrat::set_opts(use.cache = TRUE)
+    on.exit(packrat::set_opts(use.cache = FALSE))
```
kevinushey: You need to use `on.exit(..., add = TRUE)` to ensure that extra calls to `on.exit()` don't overwrite handlers attached by previous calls to `on.exit()`.

(this is silly default behavior for `on.exit()`, but unfortunately it is what it is)

tests/testthat/test-cache.R
```diff
@@ -0,0 +1,55 @@
...
+                             "lib",
+                             rv$platform,
+                             getRversion(),
+                             "oatmeal")
+
+    # After the init and before we configure a cache, the package is a
+    # non-symlink. dir.exists only tells us if the package is installed, as it
+    # reads through symlinks.
+    expect_true(dir.exists(package_dir),package_dir)
+    expect_false(packrat:::is.symlink(package_dir), package_dir)
```
kevinushey: nit: I believe `testthat` sources these files with all unexported package objects made available as though they were exported (e.g. you can just use `is.symlink()` without the `packrat` prefix).

tests/testthat/test-cache.R
```diff
@@ -0,0 +1,55 @@
...
+
+    projRoot <- cloneTestProject("healthy")
+    libRoot <- file.path(projRoot,"packrat","lib")
+    srcRoot <- file.path(projRoot,"packtar","src")
+    theCache <- tempdir()
+
+    packrat::init(projRoot, options = list(local.repos = "packages"))
+
+    rv <- R.Version()
+    package_dir <- file.path(projRoot,
```
kevinushey: I think you can use `file.path(libDir(), "oatmeal")` here.

kevinushey:
Added some comments; I'm not sure what to make of the test failures at this point though (will investigate once I'm back in to office)

aronatkins:
@kevinushey I think I've addressed all your feedback. Sadly, the Travis failures weren't magically fixed by the code shuffling!

kevinushey:
Going to merge and clean up on master. Thanks for the PR!

kevinushey:
Handled in https://github.com/rstudio/packrat/commit/8e6f4b3f1735f9a9df5c314308e2f62705faa49f.

PR #350, opened 2017-01-25
initial work on untrusted cache

kevinushey:
This PR implements an alternative cache source for untrusted client packages. The goal here is to allow users to deploy applications bundling their own package sources, but ensure that these package sources live in a separate cache from the 'global' cache (to avoid malicious users poisoning the cache).

The main way this works:

- The 'untrusted' cache lives as a separate directory inside the global cache directory,
- Hashes are produced from the source tarballs uploaded by the user,
- Otherwise, the same cache machinery stays in play.

Users of the untrusted cache will need to explicitly set the `packrat.untrusted.packages` R option, to signal to Packrat which packages that are about to be restored truly are untrusted.

cc: @trestletech. Main TODO here -- add some unit tests + confirm the 'untrusted' cache directory.

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -132,22 +132,15 @@ symlinkPackageToCache <- function(packagePath, cachedPackagePath) {
 # replacement of the cached content while overwrite=FALSE with fatal=FALSE
 # uses the cached package. Using overwrite=TRUE with fatal=TRUE will err.
 moveInstalledPackageToCache <- function(packagePath,
+                                        hash,
                                         overwrite = TRUE,
-                                        fatal = FALSE)
+                                        fatal = FALSE,
+                                        cacheDir = cacheLibDir())
 {
-  ensureDirectory(cacheLibDir())
+  ensureDirectory(cacheDir)
 
-  # TODO: this logic assumes the package folder, and package name,
-  # are the same (seems to always be true in practice, but in theory
-  # they could be different?)
-  descPath <- file.path(packagePath, "DESCRIPTION")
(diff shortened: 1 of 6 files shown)


trestletech:
> Users of the untrusted cache will need to explicitly set the packrat.untrusted.packages R option, to signal to Packrat which packages that are about to be restored truly are untrusted.

I'm not sure I follow this. At restore time you'd need to whitelist which packages you are willing to restore from the untrusted cache? 

In my mind, the whole point of the untrusted cache is that because it hashes on the contents of the bundle you wouldn't need to be more trusting about any particular package. Instead, the cache would only be used if it so happens that the machine has already built the exact tar ball that you've provided. At which point we're just short-circuiting the build process since we know the exact bits that we'd get as a result of building the provided tarball.

But I'm probably misunderstanding something?

kevinushey:
Sorry, I think I wasn't clear -- I mean, it would be the responsibility of the Connect deployment script to signal to Packrat that the package tarballs in `packrat/src` are untrusted (not the user's). Otherwise, Packrat will just stick to the default behavior, and copy all installed packages to the global cache.

Perhaps the simplest way to handle this here is to just have the Connect deployment script set an environment variable (`RSTUDIO_CONNECT`) and then just key some behaviors off of that in packrat.

trestletech:
Ohh I see. Yeah, that makes total sense. I agree that having a Boolean switch would be simpler than having to enumerate each package name. We can set an option or an env var; either is fine.

PR #358, opened 2017-02-16
Add the possibility to keep older version of packages in local repository like CRAN

Valdimus:
Hello,
I add an option to repos_upload that add the possibility to keep older version of packages in CRANlike repository like CRAN does (the "archive" option, set by default to FALSE). It's useful when you don't use bundle to keep your old packrat but only packrat.lock file. Thereby it is possible to restore the packrat with older version of your own packages stored in a private repository served by a http server.


Valdimus:
A better solution is to use [Drat](https://github.com/eddelbuettel/drat) for this feature! The function archivePackages does the job.

jeffkeller87:
@kevinushey I think this is worth reopening and adding as a feature request. Being able to archive older sources is important to those of us with private CRAN-like repositories. `drat::archivePackages(".")` works, but its help page warns of its nascency. The `packrat::repos_*` function family already does so much--would be great to add one more and have a one-stop shop.

PR #360, opened 2017-02-23
added support for pacman::p_load package loaders

xhdong-umd:
Background:  issue in [rsconnect](https://github.com/rstudio/rsconnect/issues/156) and [pacman](https://github.com/trinker/pacman/issues/82)

The issue is packrat doesn't recognize pacman package loader calls. 

First I tried the methods used in original code for checking `library`, `require`, but the `match.call` method doesn't work with `pacman::p_load` because the package name argument is not named.

Then I just scan the arguments in the pacman loader call, excludes all other arguments that are not package names, then add the package names to the environment.

In my test this code can found the package names correctly, but when I tried to deploy my shiny apps with pacman loader using my custom built packrat, there is error 

> Error: Unable to retrieve package records for the following packages:
> - 'packrat'

I suspect that to make it work with my version may not be easy, so instead of testing it completely with shiny app deployment, I'm submitting the pull request for review first.

Known limitations: `pacman::p_load_gh` is not supported yet because that is more complex.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -359,10 +359,22 @@ identifyPackagesUsed <- function(call, env) {
     return()
 
   fn <- call[[1]]
-  if (!anyOf(fn, is.character, is.symbol))
-    return()
 
   fnString <- as.character(fn)
+
+  # brute check for pacman package loader. the match.call method doesn't work with pacman p_load . Instead we just scan as.character(call). Since all other parameters in p_load are logical, we can exclude them by simple pattern matching and get the package list.
+  # put code before the :: check because pacman::p_load is often used.
+  pacmanLoaders <- c("p_load")
+  if (pacmanLoaders %in% fnString) {
+    s <- as.character(call)
+    sParas <- s[2:length(s)]
+    filterOut <- c("TRUE", "FALSE", "T", "F")
+    pkgs <- sParas[!sParas %in% filterOut]
+    lapply(pkgs, function(x) env[[x]] <- TRUE)
+  }
+
+  if (!anyOf(fn, is.character, is.symbol))
+    return()
 
   # Check for '::', ':::'
   if (fnString %in% c("::", ":::")) {


xhdong-umd:
I fixed the problem in deploying by installing `packrat` from my forked github repo. I tested with my app that used pacman loader and everything worked.

I also created a simple test file to test the changes. There are some known limitations:

supported syntax

    pacman::p_load(stringr, units, install = FALSE)
    pacman::p_load(shiny, shinydashboard)

unsupported syntax
- arguments that saved in variable. 

```
    op <- TRUE
    pacman::p_load(shinyjs, install = op)
    pkgs <- c("dplyr", "tibble")
    pacman::p_load(pkgs, character.only = TRUE)
```
- similar usage for library cannot be detected with original packrat either. I assume this kind of usage is rare and that means some more dynamical package loading, which is not suitable for shiny deployment.
```
    pkg <- "ggmap"
    library(pkg, character.only = TRUE)
```

note pacman package itself will be picked up by original code no matter what:
if used pcaman::p_load, will pick up because of ::
if used library(pacman), will pick up because of library.

xhdong-umd:
I added a Rmarkdown to explain more details, and a test file I used to test the code. The test file is just a script with different pacman syntax. Proper testthat test case can be added later if needed.

dracodoc:
Hi @jcheng5 do you have time to review the PR?

kevinushey:
Thanks for taking the time to create this PR. However, I'd prefer to avoid adding this directly to Packrat (which would imply we'd have to maintain + update it), and instead provide a mechanism for users to register their own dependency resolution routines.

xhdong-umd:
If `rsconnect` can provide such a mechanism that will be great. It's easy to just use some configuration file format like `dcf` or `yaml` to register dependencies, following existing convention in `DESCRIPTION` of package development.

stanstrup:
Would be very nice to get this merged...

Issue #153, opened 2014-10-01
Error in `snapshot()` with packages installed from external source


steromano:
I have been trying to use packages from my company internal CRAN-like repository in my packrat project. I can install fine, but when I do `packrat::snapshot()` I get the error 

```
Error: Unable to retrieve package records for the following packages: ...
```

The problem is that `inferPackageRecords` sets `source = "unknown"` for these external packages. Going through the code of `getPackageRecords` (which is throwing the error), it seems that these packages are picked up by the call to `getPackageRecordsExternalSource` but then they are dropped because they have unknown source.

To expand further, the error disappears if I explicitly hack the `DESCRIPTION` file adding `Repository:CRAN` for example, or if I call `getPackageRecords` with `fallback.ok = TRUE` (although there doesn't seem to be a way to do that from `snapshot).` In both cases the source is (incorrectly) set to "CRAN" and no error is thrown.

cronacronis:
There is a [`StatET`](http://www.walware.de/goto/opensource) project, whether for version of `rj_1.1` there is no `Repository:` in `DESCRIPTION` at all. Should packrat treat missing `Repository` with some default?

kadrach:
Has this ever been resolved? Running into this issue with drat repositories.

kevinushey:
Does the R package you're hosting on drat have a `Repository:` field in the `DESCRIPTION` file? Was the package explicitly installed from that repository?

kadrach:
These are internal packages, build with devtools, and exposed via a [drat](https://cran.rstudio.com/web/packages/drat/index.html) repository.

The `DESCRIPTION` files do not expose a `Repository`. I have no experience publishing packages on CRAN, is this line not provided by the repository?

kadrach:
`drat` does not appear to add a `Repository` field. I rather like the concept of having to specify exactly which repository to use (rather than any active CRAN-like repositories). Sounds like an option in [eddelbuettel/drat](http://github.com/eddelbuettel/drat) is the better alternative? Happy to facilitate.

jsteinhart:
I am having a similar situation, but in my case I am developing a package which uses packrat and depends on [INLA](https://bitbucket.org/hrue/r-inla), a third-party package hosted from its own (CRAN-like) repository. Therefore I believe this differs from the scenario described by @steromano and @kadrach since I have no control over the DESCRIPTION file (and therefore maybe this belongs in a separate issue...)

[...]

PR #369, opened 2017-05-06
Add failing local repository tests

colearendt:
Add packages, repo, and tests for local repository behavior.  Both tests fail per discussion in #153 

- `breakfastcereal` project is using `cereal` package 1.0.0, but the repo `testrepo` has moved on to version 2.0.0.  The `Repository: testrepo` field is set, but packrat still fails to pull sources for 1.0.0 (even though sources for that version are available in the repository).  
One note on this one... packrat.lock picked up a dependency on packrat-4.8.1, which is probably bad for go-forward testing here.  Not sure how to resolve that issue.   Making this test pass would be hugely helpful to the maintenance of external packages not yet on (or not relevant for) github or CRAN, so projects can grab sources for old versions of the package without issue.  i.e. a `drat` repo

- `breakfastyogurt` project is using the `yogurt` package 1.0.0, but the `Repository: testrepo` field is not set in the package.  Packrat fails to load the package, even though the repository is defined in `getOption('repos')`, so finding the package `yogurt` should be do-able.  Per discussion on #153 , the preferred behavior seems to be loading the package even if `Repository: testrepo` is not set.  See comment from 10/4/2016 2:04am


colearendt:
NOTE: Need to update the test for `breakfastcereal`.  Using `drat::archivePackages` allows things to work perfectly fine!

PR #386, opened 2017-06-30
Change to dir after checking name length.

jwg4:
If we try to change the working directory before checking the length of `dirName`, we don't get the right protection from problems working out the `dirName`. The call to `setwd` will fail with an argument of `character(0)`. We will get the following
```
Error in setwd(dirName) : character argument expected
```
rather than the error message "Couldn't infer top-level directory name..." as intended.


kevinushey:
Thanks!

Issue #258, opened 2015-11-18
Separate appDependencies() from init() and snapshot()


benb111:
Wrote a post about this in the packrat-discuss Google Group: https://groups.google.com/forum/#!topic/packrat-discuss/m0kKGy_W3-0.

Have you considered allowing users to init() or restore() without scanning the subdirectories for libraries to include?  That would be very useful for me, and would make it easier to work around any bugs in the scanning (which I seem to be running into when using packrat with RStudio).  As it is, if there is any issue in the auto-discovery of packages, you can't use packrat at all.  And the basic functionality of snapshotting to source and building from source doesn't seem to require this.

benb111:
To make this more concrete, I'm currently blocked from snapshotting with this error:

    Error: Unable to retrieve package records for the following packages:
    - 'GLMMGibbs', 'R6testA', 'bread', 'breakfast', 'foo', 'foo1', 'foo2', 'oatmeal', 'pkgA', 'pkgB', 'sashimi', 'testRcppClass', 'testRcppModule', 'yags'
In addition: There were 42 warnings (use warnings() to see them)

None of these are real dependencies.  For example, when I grep for oatmeal, I see

    packrat/lib/x86_64-apple-darwin14.1.0/3.1.3/packrat/tests/testthat/helper-packrat.R:  cat("library(oatmeal)", file = file.path(libraryRootDir(project), "lib-current.R"), sep = "\n")

So one problem is that packrat sees this as a dependency, but really I'd just like to disable the search for dependencies altogether.

kevinushey:
Ack, thanks. It looks like packrat is not excluding its own directory when crawling for dependencies, and this is an artefact of the fact that we now include tests in the CRAN tarball. :/

I agree it could also be useful to control whether packrat automatically infers dependencies -- this would put the onus on the user to determine what they want in their library (and want to snapshot) but I am sure is desired by more users than just yourself!

kevinushey:
Hmm, I'm a bit confused now because I can't reproduce this on OS X. Can you paste your `sessionInfo()`? It looks like we _do_ explicitly omit the `packrat/` directory; for some reason that's being missed in your case?

benb111:
Thanks for the quick response!  Here's the sessionInfo()

    > sessionInfo()
    R version 3.1.3 (2015-03-09)
    Platform: x86_64-apple-darwin14.1.0 (64-bit)
    Running under: OS X 10.10.5 (Yosemite)

    locale:
[...]

PR #397, opened 2017-09-05
Inference of dependencies optional on initialisation

itsakettle:
Hello,

This is to allow a packrat project to be initialised without searching the code for dependencies. It is useful to have more control of what package versions etc are installed when initialising packrat for a work in progress project.

I think this would close https://github.com/rstudio/packrat/issues/258.

To do this an `infer.dependencies` argument has been added to the `init` function (and also `snapshot` as this could be useful as well).

Tests all pass and an additional test has been added.

I've sanity checked by initialising a new project with the new argument and without it.

(Also, thanks for creating and maintaining packrat!)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # Packrat 0.4.9 (unreleased)
 
+- The infer.dependencies argument can now be used to switch off the scanning of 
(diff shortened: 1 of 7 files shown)


kevinushey:
Awesome! Thanks for the PR, this looks great.

PR #400, opened 2017-09-09
add "wininet" download method as default for non-secure with R>=3.2

### LINKED ISSUE #399: download method "internal" with setInternet2 defunct from R>=3.3
packrat::snapshot() fails to download source
```
> sessionInfo()
R version 3.4.1 (2017-06-30)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1
```
```
Error in getSourceForPkgRecord(pkgRecord, sourceDir, availablePkgs, repos) : 
  Failed to retrieve package sources for openssl 0.9.6 from CRAN (internet connectivity issue?) [0.9.7 is current] 

```
I stepped through the code and the pathes were fine - it attempted to download the correct archive file but eventually stopped with an error for setInternet2 in **downloadImpl**:
```
Error: 'seti2' is defunct.
See help("Defunct")
```
Apparently this is the case since R3.3. Checking the code it seems, there is a workaround with "wininet" in the secureDownloadMethod but not the downloadImpl which was used in my case because I use a http repo (as there were authorization issues on some machines).
Setting the method to "wininet" (in #400) and the issue does not appear anymore

shiosai:
fixes #399 for me

### DIFF
--- a/R/downloader.R
+++ b/R/downloader.R
@@ -207,6 +207,12 @@ inferAppropriateDownloadMethod <- function(url) {
   if (is.linux() || is.mac() || isSecureWebProtocol)
     return(secureDownloadMethod())
 
+  
+  # Use "wininet" as default for R >= 3.2
+  if (getRversion() >= "3.2")
+    return("wininet")
+  
+  # default
   return("internal")
 }
 
--- a/R/downloader.R
+++ b/R/downloader.R
@@ -209,7 +209,7 @@ inferAppropriateDownloadMethod <- function(url) {
 
   
   # Use "wininet" as default for R >= 3.2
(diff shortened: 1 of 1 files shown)


R/downloader.R
```diff
@@ -207,6 +207,12 @@ inferAppropriateDownloadMethod <- function(url) {
   if (is.linux() || is.mac() || isSecureWebProtocol)
     return(secureDownloadMethod())
 
+  
+  # Use "wininet" as default for R >= 3.2
+  if (getRversion() >= "3.2")
```
kevinushey: This check should also be guarded by `is.windows()`, I think.

PR #409, opened 2017-10-06
retry download.packages and stop if nothing is downloaded

### LINKED ISSUE #404: Packrat should log network errors more usefully
`Error in fileLoc[1, 2]: subscript out of bounds` is an error that's returned when network instability prevents `packrat::restore` from completing. It should be changed to reflect the fact that `download.packages` failed due to network instability.

aronatkins:
The `fileLoc` error was caused by a failed package download. We trapped for that failed download but only triggered a `warning` instead of `stop`. This change retries `download.packages` (modeled after `downloadWithRetries`) and stops if the retried download results in no `fileLoc` rows.

fixes #404

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Package: packrat
 Type: Package
 Title: A Dependency Management System for Projects and their R Package
     Dependencies
-Version: 0.4.8-47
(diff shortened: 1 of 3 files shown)


kevinushey:
Thanks!

PR #414, opened 2017-10-12
Tell packrat not to look for dependencies in certain directories

agwells:
This pull request adds a new option `ignore.directories`, that will tell packrat not to scan for dependencies in the R code in those directories.

My personal use-case for this, is that I have copies of some older reports in my project's git repository. They're useful for reference purposes, but I don't want their dependencies installed, and I don't want to see the warning about them being untracked every time I do `packrat::status()`.

When I was searching the Internet to see if there was an existing solution for this, I found that a second use-case is if a project has a very large amount of files in the `./data` directory, in which case it takes packrat a long time to scan them.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -122,6 +122,27 @@ dirDependencies <- function(dir) {
   packratDirRegex <- "(?:^|/)packrat"
   R_files <- grep(packratDirRegex, R_files, invert = TRUE, value = TRUE)
 
+  ## Avoid anything on the list of ignored directories
+  ignoredDir <- get_opts("ignored.directories")
+  if (length(ignoredDir) > 0) {
+    # Make sure all the directories end with a slash...
+    ignoredDir <- ifelse(
+      substr(ignoredDir, nchar(ignoredDir), nchar(ignoredDir)) != "/",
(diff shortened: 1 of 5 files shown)


R/dependencies.R
```diff
@@ -122,6 +122,27 @@ dirDependencies <- function(dir) {
...
+      substr(ignoredDir, nchar(ignoredDir), nchar(ignoredDir)) != "/",
+      paste0(ignoredDir, "/"),
+      ignoredDir
+    )
+
+    # Make a regex to match any of them.
+    ignoredDirRegex <- paste0(
+      "(?:^",
+      paste0(
+        ignoredDir,
```
kevinushey: Should we escape special characters here? (E.g. if a user wanted to ignore a directory called `data.old`, we might need to escape that as `data\.old`)

kevinushey:
LGTM (sorry for taking so long to review!)

R/dependencies.R
agwells: Good point, it should be escaped! Fortunately most of the regex special characters aren't very common in filenames, other than ".".

Without escapes, "data.old" will still match "data.old", but also incorrectly matches e.g. "dataFold"

PR #421, opened 2017-10-25
Adding docs for some undocumented options

agwells:
The "lib-r" and "lib-ext" directories can be customized following
the same pattern as the "src" and "lib" directories. But this is
not mentioned anywhere in the documentation.

Adding it to the table that describes the other path customization
options, under the "packrat-resources" help page.


agwells:
I encountered a situation where it would be helpful to be able to customize the location of the `packrat/lib-r` and `packrat/lib-ext` directories. I searched the packrat package's help, and functions list, and only saw it mentioned that I could customize the `packrat/src`, `packrat/lib`, and `packrat/bundles` directories. I was about to issue a pull request to make `lib-r` and `lib-ext` customizable as well, when I discovered that they were already customizable in the code, but it had not been mentioned anywhere in the docs. :)

Since the "packrat-resources" help page already has a table with the options to set the other paths, that seems like a logical place to put them. Since they don't have API functions to expose their current value, I've put _(none)_ in the "API" column of the table.

kevinushey:
Thanks for the PR -- LGTM!

Issue #425, opened 2017-11-08
Add a helper to change CRAN repos url in package.lock


cderv:
#378 discussion @kevinushey said about deployement of packrat project and change of CRAN mirror: 
> One way would be to modify the Packrat lockfile at deploy time, to change the repositories to whatever is desired.

It is the way we go when deploying code bundle with packrat. We often need to modify the CRAN repos indicated in `packrat.lock` to change the CRAN-like repos used when restoring. We do that in _bash_ with `sed` or in R directly in custom deployement function (wrappers that include changes in environment variable too.

Would it be be accepted to have a helper fonction in `packrat` to make this step easier when deploying code ? There are helpers for setting cran-like repo (`repo_add`, `repo_set` and friends) but not for changing `packrat.lock` without another snapshot.

Currently, I use something like this : 

```r
change_packrat_repos <- function(cran_url) {
    lockfile <- packrat:::readDcf("packrat/packrat.lock")
    lockfile[1, "Repos"] <- paste("CRAN", cran_url, sep = "=") 
    packrat:::write_dcf(lockfile, file = "packrat/packrat.lock")
}
```

There seems to be other use case for modifying `packrat.lock` repos field like #227 . 

Thanks. Happy to contribute and help.

kevinushey:
I would be open to this. Perhaps:

    packrat::set_lockfile_repos(repos)

or

    packrat::set_lockfile_metadata(repos = repos)

cderv:
Glad your are interested! 👍 
I agree with you on the second API and I will PR a first attempt soon based on your suggestion.
Thanks!

cderv:
closed by #429

PR #429, opened 2017-12-01
Set metadata in packrat lockfile

cderv:
Hey, 

This is related #425. @kevinushey I followed you advice and made one function to change different metadata. Just two for the moment `repos` and `r_version`. I think the others are not useful to modify. 

Can you take a look and tell me if the direction I took is ok ? 

I tried to write some tests but I don't manage to make them work. It seems like other test are not working either. 
I tried something like that
```r
projRoot <- cloneTestProject("healthy")
  init(enter = FALSE, projRoot, options = list(local.repos = "packages"))
``` 
but `init` does not work as expected.
help appreciated for those tests. thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/rstudio/packrat/429)

### DIFF
--- a/R/lockfile-metadata.R
+++ b/R/lockfile-metadata.R
@@ -1,0 +1,40 @@
+#' change packrat lockfile metadata manually
+#'
+#' @param repos A named character vector of the form \code{c(<repoName> = "<pathToRepo>")}
+#' @param rversion A one-legnth character vector with suitable numeric version
+#'   string. See \code{\link[base]{package_version}}
+#' @param project  The project directory. When in packrat mode, defaults to the current project;
+#'   otherwise, defaults to the current working directory.
+#' @export
+#'
+#' @examples
+set_lockfile_metadata <- function(repos = NULL, rversion = NULL, project = NULL) {
+  project <- getProjectDir(project)
+  lf_filepath <- lockFilePath(project)
+  if (!file.exists(lockFilePath)) {
+    stop(paste(lockFilePath, " is missing. Run packrat::init('",
+                 project, "') to generate it.", sep = ""))
+  }
+  lf <- as.data.frame(readDcf(lf_filepath), stringsAsFactors = F)
+
+  # update repos
+  if (!is.null(repos)) {
+    # Windows automatically transforms \n to \r\n on write through write.dcf
+    separator <- ",\n"
+    reposString <- paste(names(repos), unname(repos), sep = "=", collapse = separator)
+    lf[1, "Repos"] <- reposString
+  }
+
+  # update rversion
+  if (!is.null(rversion)) {
+    if (length(rversion) > 1) {
+      stop("RVersion metadata must contains one element only", call. = F)
+    }
+    lf[1, "RVersion"] <- as.character(package_version(rversion))
+  }
+  # write back the lockfile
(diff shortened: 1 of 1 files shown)


R/lockfile-metadata.R
```diff
@@ -0,0 +1,102 @@
...
+#' # changes R version
+#' rver <- old_rver <- get_lockfile_metadata("r_version")
+#' rver
+#' rver <- "3.4.1"
+#' set_lockfile_metadata(r_version = rver)
+#' get_lockfile_metadata("r_version")
+#' # Setting back old state
+#' # set_lockfile_metadata(r_version = old_rver)
+#' }
+set_lockfile_metadata <- function(repos = NULL, r_version = NULL, project = NULL) {
```
kevinushey: Most of the Packrat APIs take the project as the first argument -- can we follow that convention here?

R/lockfile-metadata.R
```diff
@@ -0,0 +1,102 @@
...
+  project <- getProjectDir(project)
+  lf_filepath <- lockFilePath(project)
+  if (!file.exists(lf_filepath)) {
+    stop(paste(lockFilePath, " is missing. Run packrat::init('",
+                 project, "') to generate it.", sep = ""))
+  }
+  lf <- as.data.frame(readDcf(lf_filepath), stringsAsFactors = F)
```
kevinushey: Can we use `FALSE` instead of the abbreviated `F` here?

R/lockfile-metadata.R
cderv: I followed the convention of `set_opts` and `get_opts` that I find close in functionality. `project` is last in `get_opts` and second in `set_opts`. From usage :
```r
get_opts(options = NULL, simplify = TRUE, project = NULL)
set_opts(..., project = NULL, persist = TRUE)
```
bad idea ? 
Still want me to change ?

R/lockfile-metadata.R
cderv: Absolutely right ! Good practice still not automatic for me. Done!

R/lockfile-metadata.R
kevinushey: Good point! I think you're right; let's leave it as is then.

kevinushey:
Thanks!

cderv:
Oh thanks for merging! 
I thought it needed a test file but as I did not manage to make it work I did not push it. If you want me to still try for later do not hesitate to tell me! 
Should we add a bullet to NEWS.md too ? 
I also closed manually the related issue. Forgot to add closing keyword in the PR or commit message.

thanks again for you review !

kevinushey:
I added a NEWS bullet and bumped the version here:

https://github.com/rstudio/packrat/commit/6ff1185e988331a58c04f2dd387e0a83e6355743

I'd welcome a PR adding tests as well if you have the time to put something together as well. Thanks for taking the time to put this together!

cderv:
Thanks, should have looked in the other commit.. 
For the test, I tried to get inspiration from other tests but they are skipped on cran and does not seem to work anymore. I did not manage to created a dummy packrat project in temp folder.i will keep trying and I will come up with something eventually. You will welcome a PR when it is ready. No rush but will be better if tested.

PR #432, opened 2017-12-08
restoreWithCopyFromCache to check for dir not file

javierluraschi:
cloudml is blocked from using packrat restore since I'm hitting this bug in `restoreWithCopyFromCache()` where we test for a file instead of a dir.


kevinushey:
But `file.exists()` returns `TRUE` for directories as well:

```R
> dir.create("hello")
> file.exists("hello")
[1] TRUE
```

Are you sure this is the error you're bumping into? Also, please don't merge PRs before review. I agree this change is appropriate but I'd like to fully understand why the change is required.

PR #434, opened 2017-12-17
Update README.md

kendonB:
made it clear that unbundling restores from a snapshot. Just bundling then unbundling doesn't seem to do anything across OSs.


kevinushey:
Thanks!

Issue #447, opened 2018-01-25
error when fetching package sources from private GitHub repo


ras44:
I'm using packrat and install_github to pull a specific version of a package from a private github repo:
```
install_github('<MY_USERNAME>/<MY_REPO>@v0.2.0',
               host='https://<MY_INTERNAL_GITHUB_URL>/api/v3',
               auth_token= <MY_GITHUB_PAT>
)
```

When I run 
```
packrat::snapshot()
```

I receive:
```
Adding these packages to packrat:
                          _      
    <MY_REPO>   0.2.0

Fetching sources for <MY_REPO> (0.2.0) ... Error in snapshotSources(project, activeRepos(project), allRecordsFlat) : 
  Errors occurred when fetching source files:
Error in untar2(tarfile, files, list, exdir, restore_times) : 
  incomplete block on file
```

I've checked the `packrat/` directory in the project, and it contains what looks like the repo in 
```
packrat/lib/x86_64-redhat-linux-gnu/3.4.3/<MY_REPO>
```
By that, I mean that directory contains `DESCRIPTION, INDEX, NAMESPACE, extdata, help, html, Meta, R, tests` files or directories.

On the other hand, 
```
packrat/src/<MY_REPO> 
```
is empty.

I've checked the `DESCRIPTION` file in `packrat/lib/x86_64-redhat-linux-gnu/3.4.3/<MY_REPO>` and it contains the correct URL for the repo.

I've also tried setting the GITHUB_PAT environment variable to <MY_GITHUB_PAT>, but it still fails.

From what I've read, packrat uses the URL in description to find where the file was pulled from, and should be able to work with packages that were installed via devtools::install_github.

Any suggestions for what could be going on?  Thanks in advance for your time.

kevinushey:
Is `devtools` installed and available in this project? What's the output of

```
packrat:::canUseGitHubDownloader()
```

in that project?

ras44:
Thanks for your quick response.  yes, devtools is installed and available via:
```
install.packages('devtools')
library('devtools')
```

The output of `packrat:::canUseGitHubDownloader()` is:
```
> packrat:::canUseGitHubDownloader()
[1] TRUE
```

One possibility: we are behind a firewall which requires us to specify ips for any urls accessed in `/etc/hosts`.  For example, we have entries for:

```
123.456.789.123  our.internal.github.com
123.456.789.123  codeload.our.internal.github.com
```

[...]

PR #448, opened 2018-01-26
Bugfix/snapshot private GitHub packages

### LINKED ISSUE #442: Snapshot files with package installed from private GitHub repo
I have a project that includes a package `mypkg` installed from a private GitHub repo using `install_github(, auth_token = ...)`. The installation part seemed to work fine, but when trying to generate a snapshot later, the `src/mypkg` directory is empty and I get this error:

```
Fetching sources for mypkg (0.1.1) ... Error in snapshotSources(project, activeRepos(project), allRecordsFlat) : 
  Errors occurred when fetching source files:
Error in untar2(tarfile, files, list, exdir, restore_times) : 
  incomplete block on file
```

I tried manually generating a tarball and putting it in `src/mypkg`, but I get the same error. Is this just unsupported? Is there a workaround?

ras44:
resolves #442 and #447 by removing hard-coded `api.github.com` URL in download step.  Packages installed from private github repos are not located at api.github.com.  

Instead, this generates the url from the RemoteHost variable from the DESCRIPTION file, which should specify where the package was installed from.  The RemoteHost variable is also used to generate the protocol.  The protocol choice for users using `method='internal'`, has been removed, which could cause unwanted side-effects.

Rather than using `downloadWithRetries()`, which attempts to `inferAppropriateDownloadMethod`, and was leading to `method=libcurl`, which looks like it is not supported in `canUseLibCurlDownloadMethod`, we choose to use `githubDownload()`, since we know it is a github repo.

To work, the user must set the GITHUB_PAT environment variable:
```
#set GITHUB_PAT environment variable
Sys.setenv(GITHUB_PAT='<YOUR_GITHUB_PAT')
#confirm PAT is available with devtools:github_pat(), which is used by githubDownload(...)
devtools::github_pat() 
# before installing any other packages from github.com, user must unset GITHUB_PAT
Sys.unsetenv('GITHUB_PAT')
```

I've confirmed that this works for a test case with both an internal company package and an `github.com/rstats-db/DBI`.

### DIFF
--- a/R/restore.R
+++ b/R/restore.R
@@ -202,10 +202,20 @@ getSourceForPkgRecord <- function(pkgRecord,
       error = function(e) "internal"
     )
 
-    fmt <- "%s/repos/%s/%s/tarball/%s"
-
-    archiveUrl <- sprintf(fmt, pkgRecord$remote_host, pkgRecord$gh_username, pkgRecord$gh_repo, pkgRecord$gh_sha1)
-
+    if(!exists('pkgRecord$remote_host') || is.null(pkgRecord$remote_host) || pkgRecord$remote_host == ''){
+      message('devtools version < 1.13.5, downloading from api.github.com')
+      protocol <- if (identical(method, "internal"))
+        "http"
+      else
+        "https"
+      fmt <- paste(protocol,'://api.github.com/repos/%s/%s/tarball/%s',sep='')
+      archiveUrl <- sprintf(fmt, pkgRecord$gh_username, pkgRecord$gh_repo, pkgRecord$gh_sha1)
+    }
+    else {
+      fmt <- "%s/repos/%s/%s/tarball/%s"
+      archiveUrl <- sprintf(fmt, pkgRecord$remote_host, pkgRecord$gh_username, pkgRecord$gh_repo, pkgRecord$gh_sha1)
+    }
+    
     srczip <- tempfile(fileext = '.tar.gz')
     on.exit({
       if (file.exists(srczip))
--- a/R/restore.R
+++ b/R/restore.R
@@ -202,13 +202,9 @@ getSourceForPkgRecord <- function(pkgRecord,
       error = function(e) "internal"
     )
 
-    protocol <- if (identical(method, "internal"))
-      "http"
-    else
-      "https"
+    fmt <- "%s/repos/%s/%s/tarball/%s"
 
-    fmt <- "%s://api.github.com/repos/%s/%s/tarball/%s"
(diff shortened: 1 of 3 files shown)


R/restore.R
```diff
@@ -202,22 +202,19 @@ getSourceForPkgRecord <- function(pkgRecord,
...
 
-    protocol <- if (identical(method, "internal"))
-      "http"
-    else
-      "https"
+    fmt <- "%s/repos/%s/%s/tarball/%s"
 
-    fmt <- "%s://api.github.com/repos/%s/%s/tarball/%s"
-    archiveUrl <- sprintf(fmt, protocol, pkgRecord$gh_username, pkgRecord$gh_repo, pkgRecord$gh_sha1)
+    archiveUrl <- sprintf(fmt, pkgRecord$remote_host, pkgRecord$gh_username, pkgRecord$gh_repo, pkgRecord$gh_sha1)
```
kevinushey: The `remote_host` field was added with version v1.13.5 of devtools, if I'm understanding correctly:

https://github.com/rstudio/packrat/blob/3eb9c1250ae4384012164a75282f6992c4aa972e/R/github.R#L6

That was quite a while ago, but there's a chance there's still older R installations that don't have this version installed. Can we check to see if `pkgRecord$remote_host` is `NULL` or empty, and fallback to the GitHub API URL if it's not available?

kevinushey:
This is what the lockfile entry looks like for Packrat when snapshotting on this branch:

```
Package: packrat
Source: github
Version: 0.4.8-1
GithubRepo: packrat
GithubUsername: ras44
GithubRef: bugfix/snapshot_private_github_packages
GithubSha1: 454c6c0e3ddd346c6511b645ede2e730318d5917
remote_host: https://api.github.com
remote_repo: packrat
remote_username: ras44
remote_ref: bugfix/snapshot_private_github_packages
remove_sha1: 454c6c0e3ddd346c6511b645ede2e730318d5917
Hash: 8899dbbe3edff321508a3685d1552764
```

Since we're only using the `remote_host` field, is it okay if we only keep that field? Otherwise we're duplicating a number of unneeded entries.

We also maintain a list of aliases here, for remapping of names:

https://github.com/rstudio/packrat/blob/3eb9c1250ae4384012164a75282f6992c4aa972e/R/lockfile.R#L266-L284

Can we add `remote_host` there as well?

ras44:
@kevinushey I think I've resolved all the issues.  Let me know if I missed anything- thanks

kevinushey:
Looks good! Thanks for taking the time to investigate the issue and submit the PR. I'm going to merge and then make some final tweaks.

ras44:
Thanks, Kevin- glad to help!

PR #475, opened 2018-06-11
Make Packrat secure by default

ankane:
Hey, thanks for this awesome library 👍 

This PR changes Packrat use HTTPS by default. It switches the default repo from `http://cran.rstudio.com/` to `https://cran.rstudio.com/`, as well as updates links in the doc that are also available over HTTPS.


kevinushey:
Thanks!

PR #481, opened 2018-07-06
Added support for Bitbucket

### LINKED ISSUE #304: Handle devtools::install_bitbucket
While `devtools::install_github` is nicely integrated with `packrat`, `devtools::install_bitbucket` is not. Please extend `packrat`s functionality to cater to those hosting their projects on `bitbucket` rather than `github`.

mariamedp:
Hi! I've added support for packages in Bitbucket repos (both public and private), I guess it closes #304. 

The code is the same except for some details (API calls, field names in the DESCRIPTION file, ...), so Github and Bitbucket chunks could be probably combined with some adjustments for each case. But I think it would be better to do it with a third case in mind (Gitlab?), so I've just left it separate for the moment.

About private authentication: when installling private Github packages with devtools, there is a `devtools::github_pat` function that retrieves the Personal Access Token stored in the envvar "GITHUB_PAT", but for the moment that's different with Bitbucket: there are no environment variables, and instead you set the user and password directly in the `devtools::install_bitbucket` function call. 
That doesn't fit for packrat, but it will probably [change in the future](https://github.com/r-lib/remotes/pull/104). Until then, I've created two functions `packrat:::bitbucket_user`/`packrat:::bitbucket_pwd`  that read envvars "BITBUCKET_USER" and "BITBUCKET_PWD". I think is easier to have those until the BITBUCKET_PAT envvar is properly documented.

### DIFF
--- a/R/available-updates.R
+++ b/R/available-updates.R
@@ -75,11 +75,79 @@ githubUpdates <- function(lib.loc = .libPaths()) {
   }))
 }
 
+
+bitbucketUpdates <- function(lib.loc = .libPaths()) {
+
+  do.call(rbind, enumerate(lib.loc, function(lib) {
+    pkgs <- list.files(lib, full.names = TRUE)
+    DESCRIPTIONS <- enumerate(pkgs, function(pkg) {
+      path <- file.path(pkg, "DESCRIPTION")
+      if (!file.exists(path)) return(NULL)
+      readDcf(path)
+    })
+    names(DESCRIPTIONS) <- pkgs
+    DESCRIPTIONS <-
+      Filter(function(x) "RemoteType" %in% colnames(x) && x[,"RemoteType"] == "bitbucket", DESCRIPTIONS)
+    if (!length(DESCRIPTIONS)) return(NULL)
+    if (!requireNamespace("httr")) stop("Need package 'httr' to check for Bitbucket updates")
+    do.call(rbind, enumerate(DESCRIPTIONS, function(x) {
+      url <- file.path("https://api.bitbucket.org",
+                       "2.0",
+                       "repositories",
+                       x[, "RemoteUsername"],
+                       x[, "RemoteRepo"],
+                       "refs",
+                       "branches")
+      response <- httr::GET(url)
+      status <- response$status
+      if (response$status == 403) {
+        warning("rejected by server", call. = FALSE)
+        sha <- NA
+      } else if (!response$status == 200) {
+        warning("failed to get tracking information for Bitbucket package '",
+                x[, "Package"],
+                "'; did its associated repository move?",
+                call. = FALSE)
+        sha <- NA
+      } else {
+        content <- httr::content(response, "parsed")
+        ## Find the index of the response with the appropriate name
+        index <- which(sapply(content$values, `[[`, "name") == x[, "RemoteRef"])
+        if (!length(index)) {
+          warning("no reference '", x[, "RemoteRef"],
+                  "' found associated with this repository; was the branch deleted?",
+                  call. = FALSE)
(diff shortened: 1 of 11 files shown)


kevinushey:
Thanks so much for taking the time to write a PR! Things overall look great; I just have a couple nitpicks re: typos and adjusting the naming of some of the environment variables.

I think we should also update the cache version number as the change will potentially change how existing packages are hashed.

kevinushey:
cc: @aronatkins; we're considering taking a PR in Packrat that will add support for BitBucket repositories, but this would require adding a new field to be used during hashing (https://github.com/rstudio/packrat/pull/481/files#diff-86fc2143e0466338b0d2f252e515041fR40) and so would likely necessitate a bump of the cache version.

Let me know if there are any Connect-related concerns here. Is Connect waiting on any other Packrat changes that might be worth landing before this particular change? We might also want to talk later about the feasibility of migrating an existing cache to the new version (e.g. by recomputing the hash for each package and then moving to the new cache directory)

mariamedp:
I had no particular reason for the envvars naming, so I've changed them to BITBUCKET_USERNAME and BITBUCKET_PASSWORD. I've also increased the cache version to v3. Thanks!

aronatkins:
@kevinushey Bumping the cache version invalidates all cached packages, even if they do not make use of the new `RemoteSha` field. That's really costly. Is it possible to do the following:
1. leave the cache version as `v2`
2. ensure that the computed hash is unchanged for records without `RemoteSha`

If `v3` is necessary, then we'll have to pay the cost. It would be better if hash computation could be backwards-compatible.

kevinushey:
@aronatkins I think your proposed solution will work and allow us to remain backwards-compatible with the current cache.

For what it's worth, currently, `devtools` will write out both `Github`-prefixed fields and `Remote`-prefixed fields when installing a package with `devtools::install_github()`, e.g.

```r
$ cat rmarkdown/DESCRIPTION
Package: rmarkdown
Type: Package
< ... snip ... >
RemoteType: github
RemoteHost: https://api.github.com
RemoteRepo: rmarkdown
RemoteUsername: rstudio
RemoteRef: master
RemoteSha: e9db8687cd960e9c30357477831650089bcd08e3
GithubRepo: rmarkdown
GithubUsername: rstudio
GithubRef: master
GithubSHA1: e9db8687cd960e9c30357477831650089bcd08e3
```

However, packages from Bitbucket (and I imagine other sources understood by `devtools`) only contain the `Remote` fields:

```r
$ cat packrat.test.pkg.pub/DESCRIPTION
Package: packrat.test.pkg.pub
Type: Package
Title: Test package for using packrat with bitbucket
Version: 0.1.1
Author: María Medina <@mariamedp>
Encoding: UTF-8
LazyData: true
RoxygenNote: 6.0.1
Built: R 3.5.0; ; 2018-07-12 16:13:56 UTC; unix
RemoteType: bitbucket
RemoteRepo: packrat-test-pkg
RemoteUsername: mariamedp
RemoteRef: master
RemoteSha: 05ca6fce805774309f208823f14cc59416ffe86a
```

So perhaps we can adopt a scheme where we use `GithubSHA1` if it exists, and `RemoteSha` otherwise.

I think _eventually_ `devtools` will phase out the `Github`-prefixed fields, but I'm not sure if there's an explicit timeline on that.

mariamedp:
Ok! I'm including all `Remote` fields then, I think all of them make sense: `RemoteType`, `RemoteHost`, `RemoteRepo`, `RemoteUsername`, `RemoteRef`, `RemoteSha` and `RemoteSubdir`.

kevinushey:
Awesome! Thanks @mariamedp.

PR #489, opened 2018-08-14
change with_extlib to work without packages

### LINKED ISSUE #488: with_extlib without the package?
The function name `with_extlib` suggests to me that I can execute an expression with the `extlib` library-path available to the expression. I understand that the function purpose is really "run something with an external-library *package* loaded temporarily", in which case I interpret it to really *mean* `with_extlib_package`. I'm not suggesting changing the name (likely too hard), but I would hope to be able to do something like:

```r
packrat::with_extlib(expr=packageVersion("testthat"))
```

This fails for two reasons, obviously. If I instead do

```r
packrat::with_extlib(character(0), packageVersion("testthat"))
```

I can bypass the first safeguard (`(is.character(packages)`), but unfortunately the presence of the external library is only given to the `library` call, not the expression itself.

This is a specific use-case, granted, though I can envision a more-extensive expression that relies on a package and perhaps implied dependencies. For instance, perhaps I need to access a specific function within the package but do not want to full attach it, e.g., `package::variable`, which results in the namespace being loaded but not attached to the search path.

The following code works (when I strong-arm things in `browser`).

```diff
     local({
         tryCatch({
             oldSearch <- search()
             libPaths <- .packrat_mutables$get("origLibPaths")
+            oldLibPaths <- .libPaths()
             if (!length(libPaths)) 
                 libPaths <- getDefaultLibPaths()
+            .libPaths(libPaths)
             for (package in packages) {
                 library(package, character.only = TRUE, lib.loc = libPaths, 
                   warn.conflicts = FALSE)
             }
             error <- try(res <- eval(call$expr, envir = envir), 
                 silent = TRUE)
             if (exists("res", envir = environment(), inherits = FALSE)) {
                 res
             }
             else {
                 stop(attr(error, "condition")$message, call. = FALSE)
             }
         }, finally = {
             newSearch <- search()
             for (path in setdiff(newSearch, oldSearch)) {
                 try(forceUnload(path))
             }
+            .libPaths(oldLibPaths)
         })
     })
```

The use of `lib.loc = libPaths` is not strictly required. It could be extended to be `.libPaths(c(libPaths, oldLibPaths))` if you feel the need, I don't have a strong feeling at this point.

I can submit a PR if this discussion is fruitful, what do you think of this type of change?

r2evans:
- fixes #488, "with_extlib without the package"
- included testthat test

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # Packrat 0.5.0 (UNRELEASED)
 
+- `packrat::with_extlib()` now works with no `packages=` provided;
+  both with and without this option, the new behavior is that `expr`
+  is executed in an environment where the original (not packrat)
+  library search path is in place.
+
 - A project is now only considered 'packified' if it has both a Packrat
   lockfile as well as the associated autoloader in the project `.Rprofile`.
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # Packrat 0.5.0 (UNRELEASED)
 
-- `packrat::with_extlib()` now works with no `packages=` provided;
+- `packrat::with_extlib()` now works with no `packages` provided;
   both with and without this option, the new behavior is that `expr`
   is executed in an environment where the original (not packrat)
   library search path is in place.
--- a/R/external.R
+++ b/R/external.R
@@ -26,7 +26,7 @@
 ##' with_extlib(expr = devtools::load_all("path/to/project"))
 ##' }
 ##' @export
-with_extlib <- function(packages=NULL, expr, envir = parent.frame()) {
+with_extlib <- function(packages = NULL, expr, envir = parent.frame()) {
 
   # need to force this promise now otherwise it will get evaluated
   # in the wrong context later on
--- a/R/external.R
+++ b/R/external.R
@@ -5,23 +5,34 @@
 ##' libraries; that is, the libraries that would be available upon launching a new
 ##' \R session.
 ##'
-##' @param packages A set of package names (as a character vector) to load for
-##'   the duration of evaluation of \code{expr}.
+##' @param packages An optional set of package names (as a character
+##'   vector) to load for the duration of evaluation of \code{expr}.
+##'   Whether \code{packages} is provided or \code{NULL} (the
+##'   default), \code{expr} is evaluated in an environment where the
+##'   external library path is in place, not the local (packrat)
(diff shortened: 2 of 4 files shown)


r2evans:
That failed build/test is not due to my patch, I think, since it's failing on download `texlive`.

Do you still require the 2014 contributor's agreement? (the one with `@rstudio.org`?)

R/external.R
```diff
@@ -5,23 +5,34 @@
...
+##' # since devtools requires roxygen2 >= 5.0.0 for this step, this
+##' # should fail unless roxygen2 is available in the packrat lib.loc
+##' with_extlib("devtools", load_all("path/to/project"))
+##' # this method will work given roxygen2 is installed in the
+##' # non-packrat lib.loc with devtools
+##' with_extlib(expr = devtools::load_all("path/to/project"))
 ##' }
 ##' @export
-with_extlib <- function(packages, expr, envir = parent.frame()) {
+with_extlib <- function(packages=NULL, expr, envir = parent.frame()) {
```
kevinushey: nit: prefer spaces between operators

tests/testthat/test-with_extlib.R
```diff
@@ -0,0 +1,24 @@
...
+
+test_that("with_extlib successfully works with no packages provided", {
+
+  ## Make sure packrat mode is off
+  if (packrat:::isPackratModeOn())
+    packrat::off()
+
+  orig_libs <- packrat:::getLibPaths()
+  .libPaths(c(file.path(getwd(), "packages"), orig_libs))
+  on.exit(.libPaths(orig_libs), add=TRUE)
```
kevinushey: nit: spaces

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # Packrat 0.5.0 (UNRELEASED)
 
+- `packrat::with_extlib()` now works with no `packages=` provided;
```
kevinushey: nit: spaces

kevinushey:
LGTM barring nitpicks.

Have you already signed a CLA? If not, would you be willing to send one to me (kevin <at> rstudio <dot> com)?

r2evans:
@kevinushey, any update on this PR? Did you get my CLA? (I sent it in Aug 20.)

kevinushey:
Sorry for taking so long to respond -- I did indeed receive it. Thanks!

Issue #490, opened 2018-08-14
Restore the packrat cache when it is left in a bad state


colearendt:
Connect uses the packrat cache, and we have seen a few occasions where the cached installations of a package got into a bad state.  Namely, everything but the `DESCRIPTION` file got deleted, and the `DESCRIPTION` file was left empty.  

No word yet on the root cause, but it would be nice for packrat to be able to recover from a bad installation in the cache.  The fix currently is to purge/destroy the entire folder, which will cause packrat to rebuild the cache.

PR #491, opened 2018-08-14
Add failing test for bad description file in cache

colearendt:
Related to #490 

This adds a failing test to aid in test-driven-development and to illustrate the behavior that is going awry on RStudio Connect.

@aronatkins @kevinushey FYI.  I can provide the support ticket link if it is helpful.


colearendt:
@kevinushey Should be good to merge with just the docs changes

kevinushey:
Thanks!

PR #492, opened 2018-08-14
detect and handle corrupt cache entries

### LINKED ISSUE #490: Restore the packrat cache when it is left in a bad state

kevinushey:
Fixes #490.

Effectively, packrat attempts to detect corrupt cache entries (at least, for this particular case where the DESCRIPTION file is missing or empty); when discovered, it evicts that copy from the cache and a new one will later take its place after the package is rebuilt.

cc: @colearendt

### DIFF
--- a/R/restore-routines.R
+++ b/R/restore-routines.R
@@ -3,6 +3,35 @@ isTrustedPackage <- function(package) {
   !package %in% untrusted
 }
 
+isCorruptPackageCacheEntry <- function(path) {
+
+  # if we don't have a cache entry, it's not corrupt
+  if (!file.exists(path))
+    return(FALSE)
+
+  # check for missing DESCRIPTION file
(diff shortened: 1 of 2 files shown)


colearendt:
Beautiful!  #491 includes a test for this behavior that is much more verbose than the one you constructed.  Should I strip out the test and just include the doc fixes in that PR?

kevinushey:
Ack, didn't mean to duplicate some effort there -- sorry about that! Let's just take the documentation fixes.

PR #504, opened 2018-09-28
fix packageVersion test; it may emit "smart" quotes

aronatkins:
`packrat` packrat tests failed without this change.


kevinushey:
We really shouldn't be testing the content of the error message anyhow -- e.g. in different locales the message may be translated and so different entirely.

PR #505, opened 2018-09-28
packrat:::hash has consistent output regardless of DESCRIPTION order

aronatkins:
With `packrat:::hash` and its `descLookup` argument, hashing is occasionally requested against DESCRIPTION files that are not within packages installed in an R library. `rsconnect` records description files as it creates a bundle; description information is also in the `manifest.json`.

Prior to this change, `packrat:::hash` produced different output if any field reordering occurred on the hashed description file.

DESCRIPTION file reordering initially feels unexpected, but is unavoidable when passing description data through an unordered construct, like a JSON object.

The field order used by `packrat:::hash` corresponds to common DESCRIPTION-file field order. This will help reduce the number of existing cached packages that see hash changes.

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -38,14 +38,27 @@ hash <- function(path, descLookup = installedDescLookup) {
   remote_fields <- if ("GithubSHA1" %in% names(DESCRIPTION)) {
     "GithubSHA1"
   } else {
+    # Mirror the order used by devtools when augmenting the DESCRIPTION.
     c("RemoteType", "RemoteHost", "RemoteRepo", "RemoteUsername", "RemoteRef", "RemoteSha", "RemoteSubdir")
   }
+
+  # Mirror the order of DESCRIPTION fields produced by `package.skeleton` and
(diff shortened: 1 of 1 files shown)


R/cache.R
```diff
@@ -38,14 +38,27 @@ hash <- function(path, descLookup = installedDescLookup) {
...
-  sub <- DESCRIPTION[names(DESCRIPTION) %in% fields]
+
+  # Create a "sub" data frame with a consistently ordered set of columns by
+  # first ensuring that every column has a (blank) value and then selecting
+  # with the ordered columns.
+  #
+  # This ensures that package hashing is not sensitive to DESCRIPTION field
+  # order.
+  missing <- setdiff(fields, names(DESCRIPTION))
+  DESCRIPTION[missing] <- ""
```
kevinushey: Would it be cleaner to just compute the intersection of fields and subset based on that? E.g.

```R
common <- intersect(fields, names(DESCRIPTION))
sub <- DESCRIPTION[fields]
```

PR #509, opened 2018-10-16
hashing ignores Remote* description fields when cran

aronatkins:
Client `DESCRIPTION` files may contain a number of `Remote` fields describing how the package was acquired.

```
RemoteType: cran
RemoteRepos: structure("https://mirror.las.iastate.edu/CRAN", .Names = "CRAN")
RemoteSha: 1.66.0-1
RemotePkgType: source
```

These fields are not present when `packrat` restores a package from CRAN. This causes a hashing difference between the authoring environment and the restoring environment.

This change ignores `Remote*` fields for "cran" repositories when computing hashes.

Related: https://github.com/r-lib/remotes/blob/c509c3acf1ba7231c877352953f055227871e883/R/install-remote.R#L140


aronatkins:
`devtools::install_cran("archdata")` is enough to get these fields added to the `DESCRIPTION`.

kevinushey:
LGTM -- thanks!

Issue #458, opened 2018-02-21
snapshot repeatedly knits all Rmd files


ras44:
after installing 32 libraries, I run
```
packrat::snapshot(ignore.stale = TRUE, infer.dependencies = FALSE)
```

Output indicates that it fetches all sources, but then I notice that the 'Packages' window is repeatedly refreshed before the command completes.  If I switch to the 'Files' window, I can see that each .Rmd file in the project is being knitted repeatedly (perhaps knitted once for each library?).  This leads to what seems like a very long duration for the snapshot operation.

ras44:
I can confirm this is caused by the appDependencies function in R/dependencies.R.  When I run `packrat:::appDependencies()` in the project, I see the same output.  I thought that by setting infer.dependencies=FALSE in packrat::snapshot(), it would not go through this process, but it seems like it still calls it at some point during the snapshot process.  It also calls it during an install_package() call.

I've been able to avoid repeated calls to appDependencies by setting:
```
packrat::set_opts(ignored.directories = c("data", "inst", "R"))
```
Could anyone let me know if that is the proper way to avoid searching for dependencies in the `/R` folder, without affecting any other functionality?

DaveRGP:
This is pretty annoying, any more firm official response?

ras44:
@DaveRGP after researching this by going through the code, I think this might actually be an issue with [R-Studio](https://github.com/rstudio/rstudio) and not packrat.  I've also had an issue where even just switching between different plots in the plot window causes files to knit (I know this is happening, because when I look at the my git window, which shows any new files or changes to files in realtime, I can see the knitted files being created in the background and then deleted).  I have a feeling there is some kind of very inefficient refresh code in the R Studio code that is doing this.

ras44:
I've created an issue in R-Studio here: https://github.com/rstudio/rstudio/issues/2859

ras44:
The discussion in https://github.com/rstudio/rstudio/issues/2859, mentions that this is an issue in packrat's hook function and recommends making changes to packrat, not RStudio.  I'll submit a PR that addresses the proposed modification.

ras44:
I've submitted PRs #513 and #512 as two ways to deal with the evaluation of file dependencies:
- #512 allows control only for Rmd files
- #513 is more general and allows the user to avoid any file dependency evaluation

ras44:
[...]

PR #512, opened 2018-10-31
addresses #458 slow snapshot

ras44:
This PR adds a `fileDependencies.Rmd.evaluate` option (default TRUE) to choose whether or not to recursively search for dependencies in R Markdown documents.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -208,7 +208,7 @@ stripAltEngines <- function(file, encoding) {
 
 fileDependencies.Rmd <- function(file) {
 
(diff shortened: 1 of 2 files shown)


kevinushey:
I would prefer having this as an R-level option rather than a project-level option, just because I think this will ultimately be fixed one way or another in Packrat and the option will ultimately be obviated.

kevinushey:
(that is, unless there's a strong reason why this needs to be a project option)

ras44:
sounds good, I've updated both PRs with R-level option

ras44:
closing in favor of #513

PR #513, opened 2018-10-31
opts.evaluate addresses #458 slow snapshot

ras44:
This PR adds a `fileDependencies.evaluate` option (default TRUE) to choose whether or not to recursively search for dependencies.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -144,11 +144,13 @@ dirDependencies <- function(dir) {
     R_files <- grep(ignoredDirRegex, R_files, invert = TRUE, value = TRUE)
   }
 
-  sapply(R_files, function(file) {
-    filePath <- file.path(dir, file)
-    pkgs <<- append(pkgs, fileDependencies(file.path(dir, file)))
+  if (opts$fileDependencies.evaluate) {
+    sapply(R_files, function(file) {
+      filePath <- file.path(dir, file)
+      pkgs <<- append(pkgs, fileDependencies(file.path(dir, file)))
 
-  })
+    })
+  }
 
   ## Exclude recommended packages if there is no package installed locally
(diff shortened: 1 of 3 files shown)


kevinushey:
Looks like this PR is duplicated from https://github.com/rstudio/packrat/pull/512, is that intentional?

kevinushey:
Sorry, I see the motivation now in https://github.com/rstudio/packrat/pull/513.

R/dependencies.R
```diff
@@ -144,7 +144,7 @@ dirDependencies <- function(dir) {
     R_files <- grep(ignoredDirRegex, R_files, invert = TRUE, value = TRUE)
   }
 
-  if (opts$fileDependencies.evaluate) {
```
kevinushey: What about `if (identical(getOption("packrat.dependency.discovery.enabled"), FALSE))`? IIUC this disables all of the regular dependency discovery done by Packrat.

ras44: That looks better.  What about `if (!identical(getOption("packrat.dependency.discovery.disabled"), TRUE))`?  This would allow the default case where the variable is not defined to process the dependencies- if I understand things correctly.

kevinushey: Sounds great!

ras44:
updated PR to use `packrat.dependency.discover.disabled`

kevinushey:
Thank you! The last thing I would ask -- do we already have a CLA on file? Would you mind filling out the form at https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf and sending it to us (e-mail on the form)?

ras44:
Just sent it in- thanks!

kevinushey:
Thank you!

PR #551, opened 2019-04-12
make a fileDependencies.Rmd.tangle

schloerke:
* Goal: Retrieve dependencies from each chunk independently.  
* Reason: `learnr` encourages users to provide broken code for students.  Currently since the complete tangled file can not be parsed, then no dependencies are found.
* Solution: 
  * Tangle the file
  * Split by a known comment value
  * Parse the code chunks independently
  * Combine dependencies

* Known workaround: source a separate, working R file that contains the library calls.  It seems odd to require this.

Related issue: https://github.com/rstudio/learnr/issues/209

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -669,7 +669,7 @@ fileDependencies.Rmd.tangle <- function(file, encoding = "UTF-8") {
   for(r_chunk in r_chunks) {
     if (nchar(r_chunk) == 0) next
     try(silent = TRUE, {
-      parsed <- parse(text = r_chunk, encoding = "UTF-8")
+      parsed <- parse(text = r_chunk, encoding = encoding)
       deps <- c(deps, expressionDependencies(parsed))
     })
   }
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -627,8 +627,8 @@ fileDependencies.Rmd.tangle <- function(file, encoding = "UTF-8") {
   # discovered packages
   deps <- list()
 
-  # unique key to split R code with
-  key <- paste0("###--packrat", as.numeric(Sys.time()))
+  # unique key (line) to split R code with
+  key <- paste0("###--packrat", as.integer(Sys.time()), "\n")
 
   # rudely override knitr's 'label_code' function so
   # that we can detect dependencies within inline chunks
@@ -638,7 +638,7 @@ fileDependencies.Rmd.tangle <- function(file, encoding = "UTF-8") {
     do.call("unlockBinding", list("label_code", knitr))
     assign("label_code", function(...) {
       # paste a known key to the end to split the code chunks with
-      paste0(label_code(...), key)
+      paste0(key, label_code(...))
     }, envir = knitr)
 
     on.exit({
@@ -667,6 +667,7 @@ fileDependencies.Rmd.tangle <- function(file, encoding = "UTF-8") {
   # allows for some chunks to be _broken_ but not stop retrieving dependencies
   r_chunks <- strsplit(paste0(readLines(outfile), collapse = "\n"), key)[[1]]
   for(r_chunk in r_chunks) {
+    if (nchar(r_chunk) == 0) next
     try(silent = TRUE, {
       parsed <- parse(text = r_chunk, encoding = "UTF-8")
       deps <- c(deps, expressionDependencies(parsed))
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -318,7 +318,8 @@ fileDependencies.Rmd <- function(file) {
   } else {
     warning("knitr is required to parse dependencies but is not available")
   }
-  deps
+
+  unique(deps)
 }
 
 fileDependencies.knitr <- function(...) {
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -311,28 +311,14 @@ fileDependencies.Rmd <- function(file) {
   }, add = TRUE)
 
   if (requireNamespace("knitr", quietly = TRUE)) {
-
-    tempfile <- tempfile()
-    on.exit(unlink(tempfile))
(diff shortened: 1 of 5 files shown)


R/dependencies.R
```diff
@@ -631,3 +617,61 @@ fileDependencies.Rmd.evaluate <- function(file) {
...
+  }
+
+  stripAltEngines(outfile, encoding)
+
+  # parse each r chunk independently to retrieve dependencies
+  # allows for some chunks to be _broken_ but not stop retrieving dependencies
+  r_chunks <- strsplit(paste0(readLines(outfile), collapse = "\n"), key)[[1]]
+  for(r_chunk in r_chunks) {
+    try(silent = TRUE, {
+      parsed <- parse(text = r_chunk, encoding = "UTF-8")
```
kevinushey: Should we reuse the `encoding` parameter here? Or does `knitr` enforce UTF-8 output after `purl`?

schloerke:
label code definition: https://github.com/yihui/knitr/blob/c5d1e11b90a94b6718318616bd00d45e83dd671b/R/block.R#L504-L508

schloerke:
If we can wait for a code review on my side, that be great. Thank you for
the feedback!

I’ll change the title and ping you when it’s ready. :-)

R/dependencies.R
```diff
@@ -631,3 +618,62 @@ fileDependencies.Rmd.evaluate <- function(file) {
...
+  }
+
+  # tangle out file
+  outfile <- tempfile()
+  on.exit({
+    unlink(outfile)
+  }, add = TRUE)
+
+  # attempt to tangle document with our custom hook active
+  x <- knitr::purl(file, output = outfile, quiet = TRUE, documentation = 1L)
```
schloerke: wrap in a try catch like in https://github.com/rstudio/packrat/pull/551/files#diff-a13fefafa2442d59c624b76872450cfcL318

R/dependencies.R
```diff
@@ -631,3 +618,62 @@ fileDependencies.Rmd.evaluate <- function(file) {
...
+
+
+
+fileDependencies.Rmd.tangle <- function(file, encoding = "UTF-8") {
+
+  # discovered packages
+  deps <- list()
+
+  # unique key (line) to split R code with
+  key <- paste0("###--packrat", as.integer(Sys.time()), "\n")
```
wch: The presence of `Sys.time()` suggests that time matters, but it doesn't actually matter in this case. Using something like `sample(letters, 10, replace = TRUE)` also creates a unique-ish key but doesn't suggest that time matters, so it would be a little clearer.

R/dependencies.R
```diff
@@ -631,3 +618,62 @@ fileDependencies.Rmd.evaluate <- function(file) {
...
+
+  # rudely override knitr's 'label_code' function so
+  # that we can detect dependencies within inline chunks
+  knitr <- asNamespace("knitr")
+  if (exists("label_code", envir = knitr)) {
+    label_code <- yoink("knitr", "label_code")
+    do.call("unlockBinding", list("label_code", knitr))
```
wch: Why is `do.call` needed here and below?

cc: @kevinushey

R/dependencies.R
wch: Is `x` needed?

R/dependencies.R
```diff
@@ -631,3 +618,62 @@ fileDependencies.Rmd.evaluate <- function(file) {
...
+    return(NULL)
+  }
+
+  stripAltEngines(outfile, encoding)
+
+  # parse each r chunk independently to retrieve dependencies
+  # allows for some chunks to be _broken_ but not stop retrieving dependencies
+  r_chunks <- strsplit(paste0(readLines(outfile), collapse = "\n"), key)[[1]]
+  for(r_chunk in r_chunks) {
+    if (nchar(r_chunk) == 0) next
```
wch: No need to special case the empty string, so this can be removed.

R/dependencies.R
kevinushey: Mainly to avoid `R CMD check` warnings.

R/dependencies.R
```diff
@@ -631,3 +618,62 @@ fileDependencies.Rmd.evaluate <- function(file) {
...
+
+  # discovered packages
+  deps <- list()
+
+  # unique key (line) to split R code with
+  key <- paste0("###--packrat", as.integer(Sys.time()), "\n")
+
+  # rudely override knitr's 'label_code' function so
+  # that we can detect dependencies within inline chunks
+  knitr <- asNamespace("knitr")
```
wch: In the future if this pattern is used again, it would be nice to have something like:


```R
label_code <- yoink("knitr", "label_code")
label_code_shim <- function(...) {
  # paste a known key to the end to split the code chunks with
  paste0(key, label_code(...))
}
with_shim(knitr, "label_code", label_code_shim, {
   # Code here
})

File: R/dependencies.R (hunk 2 above)
Comment (wch) [inline R/dependencies.R]: Does this need `encoding`?

File: R/dependencies.R [hunk 7]
```diff
@@ -631,3 +618,62 @@ fileDependencies.Rmd.evaluate <- function(file) {
 
   unique(unlist(deps, recursive = TRUE))
 }
+
+
+
+
+
+fileDependencies.Rmd.tangle <- function(file, encoding = "UTF-8") {
```
wch: It would be useful to have a comment about why this function exists, and how it's robust to broken chunks.

schloerke:
Submitted for QA on our side.  Will comment when all finished.  (Maybe Monday or Tuesday?)

schloerke:
@kevinushey Ready to merge on my end.  Thank you for your patience.  Please let me know if there's anything else I can do.

PR #561, opened 2019-05-31
Added code for version specific cache

### LINKED ISSUE #557: R Version Specific Packrat Cache
@kevinushey - I'm trying to understand how I can set the R_PACKRAT_CACHE_DIR environment variable to be r version specific. 

In my experience the packrat cache is r major version specific and you will get errors saying that a package has been built by a version of R with different internals if you just use one for multiple r versions on a system (system being widows for me).

I've been setting it dynamically in ".~/.Rprofile" but this is then overridden by the projects .Rprofile when I open projects using the .Rproj file. Are there any workarounds?

adamrobinson361:
To resolve #557 - @kevinushey as discussed here is a very small pull request that sets the cache to be r version specific to avoid the internals problem of restoring with different r major versions. 

Any concerns let me know - happy to build on it to more fit the style if you like.

Adam

### DIFF
--- a/R/paths.R
+++ b/R/paths.R
@@ -269,7 +269,7 @@ packratCacheVersion <- function() {
 }
 
 cacheLibDir <- function(...) {
-  file.path(appDataDir(), packratCacheVersion(), "library", ...)
+  file.path(appDataDir(), packratCacheVersion(), paste0(R.version$major, ".", R.version$minor), "library", ...)
 }
 
 untrustedCacheLibDir <- function(...) {
--- a/R/paths.R
+++ b/R/paths.R
@@ -231,11 +231,15 @@ packrat_lib <- function() {
 
 ## A location where "global" packrat data is stored, e.g. the library cache
 appDataDir <- function() {
-  packratOption(
-    "R_PACKRAT_CACHE_DIR",
-    "packrat.cache.dir",
-    defaultAppDataDir()
-  )
+
+  # Root directory
+  rootDir <- packratOption("R_PACKRAT_CACHE_DIR",
+                           "packrat.cache.dir",
+                           defaultAppDataDir())
+
+  # R Version specific sub folder
+  file.path(rootDir, getRversion())
+
 }
 
 defaultAppDataDir <- function() {
@@ -269,7 +273,7 @@ packratCacheVersion <- function() {
 }
 
 cacheLibDir <- function(...) {
-  file.path(appDataDir(), packratCacheVersion(), paste0(R.version$major, ".", R.version$minor), "library", ...)
+  file.path(appDataDir(), packratCacheVersion(), "library", ...)
 }
 
 untrustedCacheLibDir <- function(...) {


kevinushey:
cc @aronatkins just to confirm this change won't affect RSC. This PR will change the default Packrat cache directory,  but since Connect sets `R_PACKRAT_CACHE_DIR` explicitly nothing should change on the Connect side, but wanted to double-check.

R/paths.R
```diff
@@ -269,7 +269,7 @@ packratCacheVersion <- function() {
 }
 
 cacheLibDir <- function(...) {
-  file.path(appDataDir(), packratCacheVersion(), "library", ...)
+  file.path(appDataDir(), packratCacheVersion(), paste0(R.version$major, ".", R.version$minor), "library", ...)
```
kevinushey: This will set a cache using the full R version (e.g. R 3.6.0), whereas by default R shares libraries across 'patch' versions of R (e.g. using just 3.6 in the path).

Do you think separate patch versions of R should get separate caches, or does it make sense to share them?

aronatkins:
Connect does something like the following before calling out to `packrat`:

```r
packratCacheRoot <- "/the/packrat/cache"
rVersion <- getRversion()
packratCacheDir <- path.expand(file.path(packratCacheRoot, rVersion))
# [1] "/the/packrat/cache/3.5.1"
Sys.setenv(R_PACKRAT_CACHE_DIR = packratCacheDir)
```

This means that we receive `appDataDir() == '/the/packrat/cache/3.5.1'`.

```r
packrat:::cacheLibDir("httpuv")
# [1] "/the/packrat/cache/3.5.1/v2/library/httpuv"
```

This change modifies the path between `library` and `httpuv`; it means that Connect installations would have R version information in the path twice, which would invalidate the existing cache.

What about having R version information ahead of `v2`? We would update how Connect initializes `packrat` (no longer adding version information ourselves), but we would still end up with the same path.

Another option would be to change `defaultAppDataDir`; this would be invisible to Connect since we force a value with `R_PACKRAT_CACHE_DIR`.

kevinushey:
I think you're right. I like the idea of just adding the R version information so that the cache path used by default by Packrat now matches what's used currently by Connect.

kevinushey:
It looks like `appDataDir()` is only used for the Packrat cache path, so I think modifying that to include the R version is the simplest way to make this change safely. @adamrobinson361 does that sound reasonable to you as well?

R/paths.R
adamrobinson361: Interesting that rsconnect has major minor specific caches. Ive only found the problem to be major version specific but think major minor would likely mean less need for manual edits to cache more if something did go wrong. That and aligning with rsconnect sounds sensible.

adamrobinson361:
Sounds like a good solution to me. Makes more sense really to have r version first. Happy to make the edits to the appdatadir function in this pull request. Didn’t realise you had an getRversion function in the code either so will use that

adamrobinson361:
Changes made based on feedback - Its based on major minor version like RSConnect. Tried to match your convention but feel free to change anything.

If you'd like me to make any changes let me know.

kevinushey:
LGTM! @aronatkins can you sanity check once more before merge?

aronatkins:
As written, any consumer that has overridden `packrat.cache.dir` or `R_PACKRAT_CACHE_DIR` needs to make a code change (since the R version is being appended to the resolved cache directory). This needs to be announced as a breaking change since it requires packrat users to adjust.

Another option would be to change `defaultAppDataDir` - that comes into play only when no cache directory override is present.

@kevinushey - Connect is fine with either option, since can adjust how we initialize packrat when updating to this new version.

adamrobinson361:
Happy to edit the code to make use defaultAppDataDir if that would be better?

kevinushey:
I think I prefer this change, even if it might imply breaking some current usages of the `packrat.cache.dir` option. I believe it's safer if the cache directory already has the version appended, so that users can just set a single `packrat.cache.dir` option and have it work regardless of the version of R used.

Thanks for the PR! I'll add a NEWS item to make sure users are aware of the change and the next CRAN release of Packrat will have the version bumped.

Issue #559, opened 2019-05-16
Gitlab support


stanstrup:
It seems snapshot will fail if a package comes from gitlab.

```r
Error: Unable to retrieve package records for the following packages:
- "massageR"
```

I guess you don't explicitly support gitlab but is there a workaround?

The following issues about general git support are related:
https://github.com/rstudio/packrat/issues/554
https://github.com/rstudio/packrat/issues/546

kevinushey:
Unfortunately there is no workaround, but support for GitLab will be coming with [renv](https://github.com/rstudio/renv).

PR #562, opened 2019-06-06
Add gitlab

akgold:
Closes rstudio/packrat#559 by following pattern from rstudio/packrat#481. 

@kevinushey -- let me know if you have any suggestions/edits. I've tested locally and on deploy to RStudio Connect (after replacing tarball of `packrat` on Connect server) with following in RMarkdown:
```{r}
devtools::install_gitlab("alexkgold/riskgauge_pub")
riskgauge::add2(1, 1)
```

### DIFF
--- a/R/gitlab.R
+++ b/R/gitlab.R
@@ -8,35 +8,34 @@ canUseGitlabDownloader <- function() {
 
 gitlabDownload <- function(url, destfile, ...) {
   onError(1,{
-    gitlab_user    <- gitlab_user
-    gitlab_pwd     <- gitlab_pwd
     authenticate   <- yoink("httr", "authenticate")
     GET            <- yoink("httr", "GET")
     content        <- yoink("httr", "content")
 
     user <- gitlab_user(quiet = TRUE)
     pwd <- gitlab_pwd(quiet = TRUE)
-    auth <- if(!is.null(user) & !is.null(pwd)) {
+    auth <- if (!is.null(user) && !is.null(pwd)) {
       authenticate(user, pwd, type = "basic")
     } else {
       list()
     }
 
     request <- GET(url, auth)
-    if(request$status == 401) {
-      warning("Failed to download package from gitlab: not authorized. ",
+    if (request$status == 401) {
+      warning("Failed to download package from GitLab: not authorized. ",
               "Did you set GITLAB_USERNAME and GITLAB_PASSWORD env vars?",
               call. = FALSE)
       return(1)
     }
-    writeBin(content(request, "raw"), destfile)
+
+    if (request$status == 200) writeBin(content(request, "raw"), destfile)
     if (file.exists(destfile)) 0 else 1
   })
 }
 
-#' Retrieve Gitlab user.
+#' Retrieve GitLab user.
 #'
-#' A gitlab user
+#' A GitLab user
 #' Looks in env var \code{GITLAB_USERNAME}
 #'
 #' @keywords internal
@@ -45,7 +44,7 @@ gitlab_user <- function(quiet = FALSE) {
   user <- Sys.getenv("GITLAB_USERNAME")
   if (nzchar(user)) {
     if (!quiet) {
-      message("Using Gitlab username from envvar GITLAB_USERNAME")
+      message("Using GitLab username from envvar GITLAB_USERNAME")
     }
     return(user)
   }
@@ -53,9 +52,9 @@ gitlab_user <- function(quiet = FALSE) {
 }
 
 
-#' Retrieve Gitlab password
+#' Retrieve GitLab password
 #'
-#' A bitbucket password
+#' A GitLab password
 #' Looks in env var \code{GITLAB_PASSWORD}
 #'
 #' @keywords internal
@@ -64,7 +63,7 @@ gitlab_pwd <- function(quiet = FALSE) {
   pwd <- Sys.getenv("GITLAB_PASSWORD")
   if (nzchar(pwd)) {
     if (!quiet) {
-      message("Using Gitlab password from envvar GITLAB_PASSWORD")
+      message("Using GitLab password from envvar GITLAB_PASSWORD")
     }
     return(pwd)
   }
--- a/R/gitlab.R
+++ b/R/gitlab.R
@@ -1,0 +1,72 @@
+isGitlabURL <- function(url) {
+  is.string(url) && grepl("^http(?:s)?://(?:www|api).gitlab.(org|com)", url, perl = TRUE)
+}
+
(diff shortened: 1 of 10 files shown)


R/gitlab.R
```diff
@@ -0,0 +1,72 @@
+isGitlabURL <- function(url) {
+  is.string(url) && grepl("^http(?:s)?://(?:www|api).gitlab.(org|com)", url, perl = TRUE)
```
kevinushey: Is it possible for users to host their own internal GitLab servers?

(looks like it is but that's soon going away: https://about.gitlab.com/gitlab-hosted/#githost-is-shutting-down-on-june-1st-2019; just want to double-check whether we need to support those)

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+}
+
+canUseGitlabDownloader <- function() {
+  all(packageVersionInstalled(devtools = "1.9.1", httr = "1.0.0"))
+}
+
+gitlabDownload <- function(url, destfile, ...) {
+  onError(1,{
+    gitlab_user    <- gitlab_user
```
kevinushey: You can remove this line since we're just using a function defined in our own package.

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+    gitlab_pwd     <- gitlab_pwd
```
kevinushey: Same here.

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+    authenticate   <- yoink("httr", "authenticate")
+    GET            <- yoink("httr", "GET")
+    content        <- yoink("httr", "content")
+
+    user <- gitlab_user(quiet = TRUE)
+    pwd <- gitlab_pwd(quiet = TRUE)
+    auth <- if(!is.null(user) & !is.null(pwd)) {
```
kevinushey: nit: space after `if`; use `&&` for boolean checks (`&` is vectorized)

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    request <- GET(url, auth)
+    if(request$status == 401) {
+      warning("Failed to download package from gitlab: not authorized. ",
+              "Did you set GITLAB_USERNAME and GITLAB_PASSWORD env vars?",
+              call. = FALSE)
+      return(1)
+    }
+    writeBin(content(request, "raw"), destfile)
```
kevinushey: Should we double-check that we got a 200 response code before writing the request to disk?

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+#' A gitlab user
+#' Looks in env var \code{GITLAB_USERNAME}
+#'
+#' @keywords internal
+#'
+gitlab_user <- function(quiet = FALSE) {
+  user <- Sys.getenv("GITLAB_USERNAME")
+  if (nzchar(user)) {
+    if (!quiet) {
+      message("Using Gitlab username from envvar GITLAB_USERNAME")
```
kevinushey: nit: preferred capitalization is GitLab

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+      return(1)
+    }
+    writeBin(content(request, "raw"), destfile)
+    if (file.exists(destfile)) 0 else 1
+  })
+}
+
+#' Retrieve Gitlab user.
+#'
+#' A gitlab user
```
kevinushey: nit: preferred capitalization is GitLab

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+              "Did you set GITLAB_USERNAME and GITLAB_PASSWORD env vars?",
+              call. = FALSE)
+      return(1)
+    }
+    writeBin(content(request, "raw"), destfile)
+    if (file.exists(destfile)) 0 else 1
+  })
+}
+
+#' Retrieve Gitlab user.
```
kevinushey: nit: preferred capitalization is GitLab

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+    if (!quiet) {
+      message("Using Gitlab username from envvar GITLAB_USERNAME")
+    }
+    return(user)
+  }
+  return(NULL)
+}
+
+
+#' Retrieve Gitlab password
```
kevinushey: nit: preferred capitalization is GitLab

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+#'
+#' A bitbucket password
```
kevinushey: nit: bitbucket -> GitLab

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+    pwd <- gitlab_pwd(quiet = TRUE)
+    auth <- if(!is.null(user) & !is.null(pwd)) {
+      authenticate(user, pwd, type = "basic")
+    } else {
+      list()
+    }
+
+    request <- GET(url, auth)
+    if(request$status == 401) {
+      warning("Failed to download package from gitlab: not authorized. ",
```
kevinushey: nit: gitlab -> GitLab

R/gitlab.R
```diff
@@ -0,0 +1,72 @@
...
+#' A bitbucket password
+#' Looks in env var \code{GITLAB_PASSWORD}
+#'
+#' @keywords internal
+#'
+gitlab_pwd <- function(quiet = FALSE) {
+  pwd <- Sys.getenv("GITLAB_PASSWORD")
+  if (nzchar(pwd)) {
+    if (!quiet) {
+      message("Using Gitlab password from envvar GITLAB_PASSWORD")
```
kevinushey: nit: Gitlab -> GitLab

R/gitlab.R
akgold: It looks like the thing that's going away is their single-tenant hosted server, but that you'll continue to be able to host your own GitLab server on-prem or in public cloud. Any thoughts on how to check for that?

It looks like for GitHub, we only permit github.com

```
isGitHubURL <- function(url) {
  is.string(url) && grepl("^http(?:s)?://(?:www|api).github.com", url, perl = TRUE)
}
```

kevinushey: It looks like we skip the URL check in the main restore paths, e.g. for GitHub:

https://github.com/rstudio/packrat/blob/505f12aa4ef884cc7021719f4f0e4b69a0b93354/R/restore.R#L268

And you're doing the same for GitLab in the PR so I think this PR is fine as-is. Thanks!

akgold: 👍

PR #563, opened 2019-06-17
fix up API URLs when downloading from Bitbucket

kevinushey:
cc: @aronatkins.

Most of the PR is just reformatting (adding `*Impl()` methods so they're easier to debug and step into) + some brown-bag fixes for warnings popping up with R 3.6.0. The meat of the PR is this:

```
    # API URLs get recorded when packages are downloaded with devtools /
    # remotes, but Packrat just wants to use 'plain' URLs when downloading
    # package sources.
    pkgRecord$remote_host <- sub("api.bitbucket.org/2.0", "bitbucket.org", pkgRecord$remote_host, fixed = TRUE)
```

### DIFF
--- a/R/restore.R
+++ b/R/restore.R
@@ -340,6 +340,11 @@ getSourceForPkgRecord <- function(pkgRecord,
       pkgRecord$remote_host <- paste0(protocol, "://bitbucket.org")
     }
 
+    # API URLs get recorded when packages are downloaded with devtools /
+    # remotes, but Packrat just wants to use 'plain' URLs when downloading
+    # package sources.
+    pkgRecord$remote_host <- sub("api.bitbucket.org/2.0", "bitbucket.org", pkgRecord$remote_host, fixed = TRUE)
+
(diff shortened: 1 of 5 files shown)


R/restore.R
```diff
@@ -340,6 +340,11 @@ getSourceForPkgRecord <- function(pkgRecord,
       pkgRecord$remote_host <- paste0(protocol, "://bitbucket.org")
     }
 
+    # API URLs get recorded when packages are downloaded with devtools /
+    # remotes, but Packrat just wants to use 'plain' URLs when downloading
+    # package sources.
+    pkgRecord$remote_host <- sub("api.bitbucket.org/2.0", "bitbucket.org", pkgRecord$remote_host, fixed = TRUE)
```
aronatkins: This works for public bitbucket; what about on-site bitbucket?

kevinushey: That's indeed something not handled here. I think we can leave that for a separate PR.

PR #566, opened 2019-06-19
ignore remote fields when cleaning a github-sourced package record

aronatkins:
This change makes a `packrat` annotated DESCRIPTION and a `remotes` annotated DESCRIPTION more compatible; strip the `remote_*` fields before comparing package records.

### DIFF
--- a/R/pkg.R
+++ b/R/pkg.R
@@ -442,20 +442,6 @@ pick <- function(property, package, defaultValue = NA) {
   }
 }
 
-# If called without a second argument, returns a curried function. If called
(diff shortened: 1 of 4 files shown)


R/pkg.R
```diff
@@ -442,20 +442,6 @@ pick <- function(property, package, defaultValue = NA) {
   }
 }
 
-# If called without a second argument, returns a curried function. If called
-# with a second argument then it returns the package without the indicated
-# properties.
-strip <- function(properties, package) {
```
kevinushey: I will not miss this function :)

PR #569, opened 2019-08-01
Allow empty repository field

kenahoo:
Addresses issue #528

PR #580, opened 2019-10-21
optional fields are not in package record when NULL; avoid identical in RemoteType compares

aronatkins:
The commit https://github.com/rstudio/packrat/commit/e329cf15ddfaed7786d2b6831d749d87c66eec2c from https://github.com/rstudio/packrat/pull/569 created a function to help construct package records. Unfortunately, those new records included fields that are sometimes NULL and should be omitted from the package record when NULL. Reverted the constructor for now.

The commit https://github.com/rstudio/packrat/commit/29a4fdcb5d34113222b73e4eb2affd0e3fc3f1fd started using `identical` to compare the `RemoteType` field. This will not compare if the incoming data frame was read from DCF with strings-as-factors.

The test failures on `master` were caused by these commits. This PR also includes a couple extra checks that would have helped identify the problems.

### DIFF
--- a/R/pkg.R
+++ b/R/pkg.R
@@ -283,6 +283,12 @@ getPackageRecords <- function(pkgNames,
   allRecords
 }
 
+# Return TRUE when the data frame for this package has the given RemoteType.
+hasRemoteType <- function(df, remoteType) {
+  # Do not compare with 'identical'; RemoteType may be a factor.
+  return (!is.null(df$RemoteType) && df$RemoteType == remoteType)
+}
+
 # Reads a description file and attempts to infer where the package came from.
 # Currently works only for packages installed from CRAN or from GitHub/Bitbucket/Gitlab using
 # devtools 1.4 or later.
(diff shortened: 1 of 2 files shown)


R/pkg.R
```diff
@@ -283,66 +283,99 @@ getPackageRecords <- function(pkgNames,
   allRecords
 }
 
+# Return TRUE when the data frame for this package has the given RemoteType.
+hasRemoteType <- function(df, remoteType) {
+  # Do not compare with 'identical'; RemoteType may be a factor.
```
kevinushey: Agh. I'd prefer to know where this is happening so that we can fix the problem at the source, but this should still be okay.

tests/testthat/test-packrat.R
```diff
@@ -279,12 +279,16 @@ withTestContext({
 
     # validate the installed package has properly annotated DESCRIPTION
     descpath <- file.path(libDir(projRoot), "falsy/DESCRIPTION")
-    desc <- as.data.frame(readDcf(descpath), stringsAsFactors = FALSE)
+    desc <- as.data.frame(readDcf(descpath))
```
kevinushey: Is the removal of `stringsAsFactors = FALSE` here just to ensure that we're resilient against both strings + factor fields read from the DESCRIPTION?

aronatkins: Yes. I could make these an explicit `stringsAsFactors = TRUE` if that's better.

kevinushey: It is (this is clear from the PR description.)

R/pkg.R
aronatkins: Would you rather I use the null/equality checks inline at each call-point? (reverting most of your https://github.com/rstudio/packrat/commit/29a4fdcb5d34113222b73e4eb2affd0e3fc3f1fd commit)

kevinushey: No, I think this is fine. Thanks!

PR #581, opened 2019-10-22
Updates documentation to work on Windows

bigfatbird:
Fixed a bug where it makes it impossible to run this script on a windows machine.


kevinushey:
Thank you for catching that!

bigfatbird:
I found several other things not working for me. The tutorial seems kinda outdated or not detailed enough. 

I'm trying to get a packrat workflow established at work and have a lot of problems.

With the upcoming renv in development I should ask myself though, if I should continue to work with packrat or switch over already. :D

kevinushey:
I would encourage you to give `renv` a try. Its API is largely compatible with Packrat, and all active development in the future will be focused on `renv`. Packrat will continue to be maintained and we'll provide high-priority bug fixes when possible, but most of the future work will center on `renv`.

Issue #532, opened 2019-01-28
snapshot dependency resolution behavior causes slowness


dpastoor:
@trestletech as we spoke about briefly at Rstudio Conf, here is an example of looking at the flamegraphs and data associated with a snapshot for a pretty vanilla data science project with shiny (eg tidyverse + shiny and a couple helper packages)

You can see that the snapshot call takes almost 2 minutes:

![image](https://user-images.githubusercontent.com/3196313/51809216-5a66a700-226c-11e9-9fe9-1dd61fa7da87.png)

Ironically, the callHook, 49 of the 50 seconds is likewise stuck on another snapshotImpl call.

![image](https://user-images.githubusercontent.com/3196313/51809221-74a08500-226c-11e9-8e07-6a67057e5105.png)

So nearly 97 of 99 seconds is taken just doing dependency reconciliation. 

If we drill further into the snapshotImpl method, we can also see that the cause is a recursive call for getting package dependencies, which I believe during unraveling is re-solving the dependency graph and reading the lockfile many times.

![image](https://user-images.githubusercontent.com/3196313/51809249-f0023680-226c-11e9-8553-91b86ce0cc4a.png)

This is also very evident from looking at how spiky the flame graph is

![image](https://user-images.githubusercontent.com/3196313/51809287-62731680-226d-11e9-8d26-e67af416d97e.png)

@slopp this is also why we struggle with using programmatic deployment from rsconnect and friends using the traditional bundling, as in https://github.com/rstudio/rsconnect/blob/master/R/bundle.R#L780 can take so long.

I would think by using either memoization or a separate environment as a hash table of sorts to check against before solving for a particular package would be very helpful.

IMO, this should not take more than a second or two max, which I lightly confirmed with an extremely hacky script that replicated the general gist of what is going on during the getPackageRecords call. I'm happy to help get further along with this, but figured it might be worth having a dialog around whether its worth it or if it makes sense to channel these optimization efforts elsewhere.

dpastoor:
In digging around this looks like it aligned with the findings at https://github.com/rstudio/packrat/issues/347

aronatkins:
@dpastoor - the changes in https://github.com/rstudio/packrat/pull/615 dramatically reduced the cost of `getPackageRecords`. Those changes were included in the packrat 0.6.0 release.

Are you able to re-test the performance of dependency resolution using the current CRAN release?

dpastoor:
Hi Aron,

[...]

Issue #290, opened 2016-04-06
packrat::on() very slow (5+ minutes)


raubreywhite:
When I turn packrat on, the command runs very quickly, however, afterwards RStudio will not accept any input for at least 5 minutes while it processes and updates the "packages" tab.

kevinushey:
You mentioned in a separate post that you were running on a networked drive -- I wonder if the dependency detection code, or something else, is taking an inordinate amount of time?

You might want to try disabling the Package pane:

<img width="585" alt="screen shot 2016-04-06 at 11 25 26 am" src="https://cloud.githubusercontent.com/assets/1976582/14328161/6691c0c4-fbea-11e5-834e-5a7d88d7b280.png">

raubreywhite:
Thanks for the suggestion, I just tried it and it didn't make any difference.

I did find something *extremely* strange though.

If I run packrat::on() while in RStudio project mode, it will hang for ~5 minutes after the command and refuse to take any input. If I run the same source code outside of RStudio project mode, it will only hang for ~5 seconds.

I then went into my package options and tried:
- disabling auto.snapshot
- disabling packrat

Neither of those made it any faster in project mode. 

So I'm not quite sure what is happening, but since it works well outside of project mode I will just work in that environment :-)

raubreywhite:
The above also applies when I use "packrat::on(auto.snapshot=FALSE)"

raubreywhite:
I also walked through the source code of packrat::on() and ran it line-by-line to see where it hung.

In project mode, everything works perfectly well until I hit the line

packrat:::setLibPaths(c(localLib, packrat:::libExtDir(project)))

(in function packrat:::afterPackratModeOn)

Then RStudio hangs for 5 minutes and won't respond.

I also tried restarting my R session, entering project mode and then immediately running 
a) .libPaths("F:/_SM/SMAO/Faggrupper/SMX/Sykdommer/Hepatitt C/Burden of disease/Hinta_graphs_SSI/packrat/lib")
b) packrat:::setLibPaths("F:/_SM/SMAO/Faggrupper/SMX/Sykdommer/Hepatitt C/Burden of disease/Hinta_graphs_SSI/packrat/lib")

Neither of these caused RStudio to hang at all.

raubreywhite:
Cross posting from https://support.rstudio.com/hc/en-us/community/posts/210311327-packrat-on-very-slow-5-minutes-

###########

Hi Kevin,



Thanks for the suggestion/reply.



[...]

PR #603, opened 2020-06-10
"Non recursive" optimisation - 10x speed improvement for our RStudio Connect deployments

slodge:
This PR introduces a global base R level option to change the recursive search option used by `.snapshotImpl` when calling `getPackageRecords` for inferred packages.

This PR is inspired by a previous commit https://github.com/rstudio/packrat/commit/ac1a18cf856f440b64c5db44401c350b16b94636 referenced by @ras44 in #347 - but then reversed as "it broke a couple internal RStudio applications" (@kevinushey)

Testing locally this has significantly reduced `rsconnect::writeManifest` times for our Shiny apps while producing the same manifest results.

- Performance was measured upon a single file App.R reference shiny, tidyverse and some C19 test data (see code below)
- With this change the rsconnect write manifest time was reduced 
  - from ~100 seconds 
  - to <10 seconds 

I'm not expecting this PR to be approved "as is" - while it's worked for all the apps I've tested with, I expect it may have implications in other environnments.

I'm also not glued to this change being introduced as a global R option - we just went that route as it required a minimal change and as it's easy to turn on/off during testing.

I'd be happy to consider alternative fixes and to put work into investigating them - but currently we need to do something as the RStudio Connect deployment experience is far too slow for our users. An optimisation like this one would make a significant difference to the perceived usability of our RStudio Connect instances.

I did try the other fix suggested in #347 - using `packrat::set_opts(ignored.directories=c("all","my","R","src","directories")` - however, this didn't yield any performance improvements for me even when modifying the rsconnect package to use different `snapshotImpl` arguments.

Referencing:
#347
#532

Possibly also linked to other speed issues - e.g. #290

Measurement script:
```
options(packrat.RecursiveInference = T)
tictoc::tic()
rsconnect::writeManifest(appDir="../C19", appFiles = "app.R" )
tictoc::toc(log = T)

options(packrat.RecursiveInference = F)
tictoc::tic()
rsconnect::writeManifest(appDir="../C19", appFiles = "app.R" )
tictoc::toc(log = T)
```

Test app.R
```
library(shiny)
library(tidyverse)
library(coronavirus)

all_countries <- coronavirus %>%
  pluck("country") %>%
  unique() 


ui <- fluidPage(
  
  titlePanel("C19"),
  
  sidebarLayout(
    sidebarPanel(
      selectInput(
        "country",
        label = "Which Country?", 
        choices =  all_countries, 
        selected = "United Kingdom",
        multiple = TRUE,
        selectize = TRUE)
    ),
    
    mainPanel(
      plotOutput("thePlot")
    )
  )
)

server <- function(input, output) {
  
  output$thePlot <- 
    renderPlot({
      
      coronavirus %>% 
        filter(type == "confirmed",
               country %in% input$country) %>%
        ggplot(aes(date, cases, colour=country)) +
        geom_line() +
        facet_wrap(~country)
      
  })
}

# runApp()
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Package: packrat
 Type: Package
 Title: A Dependency Management System for Projects and their R Package
     Dependencies
-Version: 0.5.0-26
+Version: 0.5.0-27
 Author: Kevin Ushey, Jonathan McPherson, Joe Cheng, Aron Atkins, JJ Allaire
 Maintainer: Kevin Ushey <kevin@rstudio.com>
 Description: Manage the R packages your project depends
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -194,7 +194,8 @@ snapshot <- function(project = NULL,
                                           project = project,
                                           available = available,
                                           check.lockfile = TRUE,
-                                          fallback.ok = fallback.ok)
+                                          fallback.ok = fallback.ok,
+                                          recursive = getOption("packrat.RecursiveInference", default = T))
 
   allRecords <- c(
     libPkgRecords,


rmccombie:
Feedback: 55750

aronatkins:
Also related is our more recent work in https://github.com/rstudio/packrat/pull/615, where packrat avoids repeatedly processing the same dependency.

ras44:
@kevinushey @aronatkins we've heard recommendations on moving towards Renv.  Do you have any thoughts on whether these types of improvements will pass through to Renv?  I'm a bit unclear if Renv is a layer over packrat, or something altogether separate.  Thanks!

kevinushey:
`renv` is effectively a ground-up rewrite of Packrat, but done in a way to be mostly compatible with Packrat, or at least the API should feel familiar to those experienced with Packrat. We plan to continue maintaining Packrat for critical fixes required for RStudio Connect, but future development will focus on the `renv` package.

In the future, we aim to eventually migrate the deployment process to use `renv` rather than Packrat, but that work has not yet been scheduled.

ras44:
@kevinushey thanks for your quick response.  So would it be safe to say that the `renv` rewrite doesn't suffer from the performance issues that this PR and others referenced appear to be trying to resolve in packrat?  Just trying to gauge whether we should switch to `renv` for superior performance, or continue using packrat but with these fixes.  Thanks for your guidance!

kevinushey:
That's correct. In general, you can view `renv` as a better Packrat. (If there's something that Packrat does that `renv` does not that you require, please let me know as well)

PR #605, opened 2020-06-23
updated GA tracking code to include new RStudio websites

sarahemlin:
Updated GA tracking code to include both new RStudio websites and sites omitted in the April change.
@kevinushey  please review


kevinushey:
Thanks!

PR #614, opened 2020-11-24
Optimise package search tree and insert verbose logger messages

andrie:
Packrat searches for recursive dependencies in the getPackageRecords() function.

However, if a package appears multiple times in the search tree, this recursion will run multiple times.

This PR optimizes the search tree by memorizing whether a package was searched before, and if so, skips the search the next time.

For my test case, this reduces the `packrat` snapshot time from 71 sec to 14 sec.

### DIFF
--- a/R/pkg.R
+++ b/R/pkg.R
@@ -184,10 +184,17 @@ getPackageRecords <- function(pkgNames,
                               lib.loc = NULL,
                               missing.package = error_not_installed,
                               check.lockfile = FALSE,
-                              fallback.ok = FALSE)
+                              fallback.ok = FALSE,
+                              .recursion.level = 1,
+                              .visited.packages = new.env(parent = emptyenv()),
+                              verbose = FALSE)
 {
   project <- getProjectDir(project)
   local.repos <- get_opts("local.repos", project = project)
+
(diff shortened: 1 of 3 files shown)


R/pkg.R
```diff
@@ -257,13 +264,30 @@ getPackageRecords <- function(pkgNames,
...
+        logger(sprintf("- getting dependency %3i of %3i %s", .iii, .nnn, record$name))
+        .iii <<- .iii + 1
+      }
       deps <- getPackageDependencies(pkgs = record$name,
                                      lib.loc = lib.loc,
                                      available.packages = available)
-      if (!is.null(deps)) {
+
+      new  <- setdiff(deps, ls(envir = .visited.packages))
+      for (pkg in new) {
```
aronatkins, line 285: We probably should update `.visited.package` based on the incoming set of packages (`pkgNames`). The approach here can still visit a package repeatedly (at most twice).

Filtering previously visited packages (the `setdiff` call) can still remain here (but could also shift to the start of the function).

aronatkins:
Replaced by https://github.com/rstudio/packrat/pull/615 (now merged).

PR #615, opened 2020-11-24
avoid repeated package dependency processing

aronatkins:
an alternate tree-pruning approach from #614; part of an ongoing discussion with @andrie

### DIFF
--- a/R/pkg.R
+++ b/R/pkg.R
@@ -195,13 +195,12 @@ getPackageRecords <- function(pkgNames,
 
   # Prior recursive steps may have already computed this package record and
   # its recursive dependencies. Avoid constructing this package record.
-  priorPkgRecords <- c()
-  for (pkgName in pkgNames) {
-    if (exists(pkgName, envir = .visited.packages)) {
-      append(priorPkgRecords, get(pkgName, envir = .visited.packages))
-    }
+  priorPkgRecords <- dropNull(lapply(pkgNames, function(pkgName) {
+    get0(pkgName, envir = .visited.packages, ifnotfound = NULL)
+  }))
+  if (length(priorPkgRecords)) {
+    pkgNames <- setdiff(pkgNames, sapply(priorPkgRecords, "[[", "name"))
   }
-  pkgNames <- setdiff(pkgNames, sapply(priorPkgRecords, "[[", "name"))
 
   if (check.lockfile) {
     lockfilePkgRecords <- getPackageRecordsLockfile(pkgNames, project = project)
--- a/R/pkg.R
+++ b/R/pkg.R
@@ -279,7 +279,7 @@ getPackageRecords <- function(pkgNames,
           missing.package = missing.package,
           check.lockfile = check.lockfile,
           fallback.ok = fallback.ok,
-          .visited.packages
+          .visited.packages = .visited.packages
         )
       }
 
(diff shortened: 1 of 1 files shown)


R/pkg.R
```diff
@@ -272,7 +278,8 @@ getPackageRecords <- function(pkgNames,
           lib.loc = lib.loc,
           missing.package = missing.package,
           check.lockfile = check.lockfile,
-          fallback.ok = fallback.ok
+          fallback.ok = fallback.ok,
+          .visited.packages
```
kevinushey, line 282: nit: pass along as named argument here, just to be safe?

aronatkins: thanks. i was rushing. :)

kevinushey:
This PR looks good to me, although I think we might need to narrow the scope a bit more since (if I understand the test failures correctly) some APIs do rely on us building the full dependency tree for packages.

What if we instead hid this behavior behind an option, e.g. something like `options(packrat.PruneDependencyTree = TRUE)`, and set that for `rsconnect`'s usages of Packrat?

aronatkins:
Updated the approach to prune the search while still producing the full dependency tree.

R/pkg.R
```diff
@@ -184,14 +184,25 @@ getPackageRecords <- function(pkgNames,
...
 
   # screen out empty package names that might have snuck in
   pkgNames <- setdiff(pkgNames, "")
 
+  # Prior recursive steps may have already computed this package record and
+  # its recursive dependencies. Avoid constructing this package record.
+  priorPkgRecords <- c()
+  for (pkgName in pkgNames) {
+    if (exists(pkgName, envir = .visited.packages)) {
+      append(priorPkgRecords, get(pkgName, envir = .visited.packages))
```
kevinushey, line 201:
```suggestion
      priorPkgRecords <- append(priorPkgRecords, get(pkgName, envir = .visited.packages))
```

As is, this statement won't update `priorPkgRecords` -- is that intentional?

R/pkg.R
```diff
@@ -195,13 +195,12 @@ getPackageRecords <- function(pkgNames,
 
   # Prior recursive steps may have already computed this package record and
   # its recursive dependencies. Avoid constructing this package record.
-  priorPkgRecords <- c()
-  for (pkgName in pkgNames) {
-    if (exists(pkgName, envir = .visited.packages)) {
-      append(priorPkgRecords, get(pkgName, envir = .visited.packages))
-    }
+  priorPkgRecords <- dropNull(lapply(pkgNames, function(pkgName) {
+    get0(pkgName, envir = .visited.packages, ifnotfound = NULL)
```
kevinushey, line 199:
```suggestion
    .visited.packages[[pkgName]]
```

as `get0` was introduced with R 3.2.0, and so (in theory) could be too recent (RStudio still declares R 3.0.1 as the minimum required version; not sure what the minimum for RSC is now.)

aronatkins: RSC declares R 3.1 as its minimum; I've changed this to use `exists` and `get`.

kevinushey:
LGTM! (Remaining test failures are due to dependent R packages not being available for older R.)

kevinushey (approved):
LGTM -- thanks for putting this together! (And thanks to @andrie for getting this work started as well.)

PR #618, opened 2021-01-27
NEWS about dependency improvement

aronatkins:
Noticed that we did not add NEWS for the recursion improvement. Here's one option for wording.


kevinushey:
Thanks!

PR #619, opened 2021-01-27
requireNamespace and loadNamespace do not accept symbols as package names

### LINKED ISSUE #602: Bug: error if requireNamespace is used in function
I am been redirected from the rsconnect [issue](https://github.com/rstudio/rsconnect/issues/423) to here. 

First, if I have code like this in my function, packrat treat the variable `pkg` as a real package name. 
```r
missing_pkgs <- lapply(packages, function(pkg) {
        if (!requireNamespace(pkg, quietly = TRUE)) pkg
})
```
Second, this code is used for checking the user environment, I don't want to install or attach the package, so that's why I use `requireNamespace` instead of `require`. packrat also thinks anything in `requireNamespace`  is needed, but they are not.

if you run `packrat:::appDependencies()`:
```r
Warning message:
In FUN(X[[i]], ...) : Package 'pkg' not available in repository or locally
```
This is a warning here, but will become an error in rsconnect and prevent my deployment. 

Is there an official way to fix this instead of my hack? 

Thanks!

aronatkins:
fixes #602

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,9 @@
   
 - Improve performance of dependency processing. (#615)
 
+- Infer package dependencies from `requireNamespace()` and `loadNamespace()`
+  only when the package name is character input. (#602)
+
 # Packrat 0.5.0
 
 - Packrat now supports both of BiocManager and BiocInstaller (as used for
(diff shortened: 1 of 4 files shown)


kevinushey:
LGTM! Associated Travis failure appears to be because many of the latest R packages on CRAN no longer support R 3.2.5, so that should be fixed separately.

PR #623, opened 2021-02-18
define setNames ahead of options.R

aronatkins:
The utils.R definition of setNames was not visible to the package-build-time
calls to setNames in options.R because of R file-load order. As a result,
packrat would fall back to stats::setNames or err if stats was not loaded.


kevinushey (approved):
LGTM!

PR #626, opened 2021-03-26
stop requiring devtools for SCM restores

aronatkins:
Playing with this idea as we're debugging an issue when downloading private GitHub packages. Untested at this point.


aronatkins:
replaced by https://github.com/rstudio/packrat/pull/649.

PR #638, opened 2021-08-13
skip_on_ci as with travis

aronatkins:


aronatkins:
A run of this branch against rhub passes, but in GHA, we're seeing

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure (test-dependencies.R:58:3): dependencies are discovered in inline R code ──
"emo" %in% packrat:::fileDependencies(emojiRmd) is not TRUE

`actual`:   FALSE
`expected`: TRUE 
── Failure (test-rmarkdown.R:22:3): We can discover dependencies with an evaluate hook ──
`deps` not equal to c("abc", "def", "ghi", "jkl").
target is NULL, current is character
```

aronatkins:
After adding pandoc to the GHA workflow, CI is passing. Also added multiple R versions, as it was unclear at first if the problems were with a version of R different than my development environment.

Issue #644, opened 2021-11-04
HTML/extras created on-disk during file-dependency probing


aronatkins:
1. Create a new directory/project.
2. Populate that directory

    ```r
    bookdown::create_bs4_book(".")
    ````

2. Add a code-block to `index.Rmd` so package dependencies are discovered.

    ````markdown
    ```{r, include = FALSE}
    library(downlit)
    library(bslib)
    library(xml2)
    ```
    ````
3. Render that book locally.

    ```r
    rmarkdown::render_site(".")
    ```

4. Attempt to deploy that directory.

    ```r
    rsconnect::deploySite(
        siteDir = ".", 
        siteName = "BOOK-TITLE", 
        account = "ACCOUNT_NAME", 
        server = "SERVER_NAME", 
        render = "server")
    ```

The deployed bundle contains a `libs` directory and `*.html` files that were created as a side-effect of running the packrat snapshot against the directory. These files cause downstream rendering problems.

These files are created by the call to `rmarkdown::render`:
https://github.com/rstudio/packrat/blob/3df1d62f77e98f846b11636b4a3203b925ec1028/R/dependencies.R#L665

aronatkins:
The eventual deployment fails with the rendering error:

```
Tweaking _book/404.html
Error in xml_children(x)[[search]] : subscript out of bounds
Calls: local ... bs4_chapter_tweak -> tweak_navbar -> template_link_icon -> <Anonymous>
Execution halted
Error: bookdown::render_book() failed to render the output format 'bookdown::bs4_book'.
```

The `404.html` file is created as a side-effect of the deploy-time `rmarkdown::render` call.

aronatkins:
Resolved with #647; NEWS appearing shortly.

PR #645, opened 2021-11-04
initial forklift of renv dependency scanning

aronatkins:
Incomplete in its current form. This was enough to prove that we can use the renv dependency analysis and avoid the problems identified in https://github.com/rstudio/packrat/issues/644.

Files copied from renv commit https://github.com/rstudio/renv/commit/56bf09bd2ad3abf5b0d28ba6abc7f5a7b9a88f01 (0.14.0-10)


aronatkins:
We're trying a different approach in #646

PR #649, opened 2021-12-20
stop using devtools::github_pat, as it is no longer available

### LINKED ISSUE #651: devtools::github_pat is gone
With the devtools-2.4.3 release, `github_pat` is gone.

https://github.com/cran/devtools/commit/d1b8fc691f12d7f640533fdf1ace30a4227287c3#diff-b6b6854a02ae177f8860f49654ff250c1554d021ae434b6efdbe99d00bdc6055

Packrat needs to stop attempting to use this function when performing an authenticated GitHub download.

aronatkins:
We have started seeing errors when folks are using packrat to install private GitHub packages. This is rooted in the fact that `devtools` has removed the `github_pat` function in its 2.4.3 release.

* use authenticated downloaders when httr is installed and
  authentication is available.
* err in github/gitlab/bitbucket downloaders, which provides
  more information on failure to restore.
* remove github/gitlab/bitbucket download calls from downloadImpl,
  as all downloadWithRetries calls are prefaced by repo-specific
  download calls.
* err on all non-200 HTTP statuses for github/gitlab/bitbucket.
  previously, whatever content we downloaded would be used,
  which could be empty, or a JSON error document, or whatever.

R errors such as the lack of `devtools::github_pat` would now appear like:

```
Error in value[[3L]](cond) : Failed to download package from URL:
- 'https://api.github.com/repos/rstudio/academyapi/tarball/6e3b86420e053359774468b364fe7f73c53f3f1b'
- Reason: Error: GitHub request failed with Error in get(name, envir = asNamespace(pkg), inherits = FALSE): object 'github_pat' not found
```

HTTP errors would now appear like:

```
Error in value[[3L]](cond) : Failed to download package from URL:
- 'https://api.github.com/repos/rstudio/academyapi/tarball/6e3b86420e053359774468b364fe7f73c53f3f1b'
- Reason: Error: GitHub request failed with Error: Unable to download package from GitHub; check the GITHUB_PAT environment variable: Client error: (404) Not Found
```

Fixes #651

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 - Stop using `devtools::github_pat` to obtain an (optional) `GITHUB_PAT`
   environment variable value. The `github_pat` function was removed in
-  devtools-2.4.3. (#651)
+  `devtools-2.4.3`. (#651)
 
 - Use authenticated downloaders for GitHub, GitLab, and Bitbucket when the
   `httr` package is installed and the appropriate credentials are available.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,14 @@
 # Packrat 0.8.0 (UNRELEASED)
 
+- Stop using `devtools::github_pat` to obtain an (optional) `GITHUB_PAT`
+  environment variable value. The `github_pat` function was removed in
+  devtools-2.4.3.
+
+- Use authenticated downloaders for GitHub, GitLab, and Bitbucket when the
+  `httr` package is installed and the appropriate credentials are available.
+
+- Provide more information about authenticated download failures.
+
 
 # Packrat 0.7.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 - Stop using `devtools::github_pat` to obtain an (optional) `GITHUB_PAT`
   environment variable value. The `github_pat` function was removed in
-  devtools-2.4.3.
+  devtools-2.4.3. (#651)
 
 - Use authenticated downloaders for GitHub, GitLab, and Bitbucket when the
   `httr` package is installed and the appropriate credentials are available.
--- a/R/bitbucket.R
+++ b/R/bitbucket.R
@@ -3,8 +3,8 @@ isBitbucketURL <- function(url) {
 }
 
 canUseBitbucketDownloader <- function() {
-  (all(packageVersionInstalled(httr = "1.0.0")) &
-     !is.null(bitbucket_user(quiet = TRUE)) &
+  (all(packageVersionInstalled(httr = "1.0.0")) &&
+     !is.null(bitbucket_user(quiet = TRUE)) &&
      !is.null(bitbucket_pwd(quiet = TRUE)))
 }
 
@@ -12,7 +12,7 @@ bitbucketDownload <- function(url, destfile, ...) {
   tryCatch(
     bitbucketDownloadImpl(url, destfile, ...),
     error = function(e) {
-      stop(sprintf("Bitbucket request failed with %s", e), call. = FALSE)
+      stop(sprintf("Bitbucket request failed: %s", e), call. = FALSE)
     })
 }
 
@@ -33,7 +33,7 @@ bitbucketDownloadImpl <- function(url, destfile, ...) {
   if (result$status != 200) {
     stop(
       sprintf(
-        "Unable to download package from Bitbucket; check the BITBICKET_USERNAME and BITBUCKET_PASSWORD environment variables: %s",
+        "Unable to download package from Bitbucket; check the BITBUCKET_USERNAME and BITBUCKET_PASSWORD environment variables: %s",
         httr::http_status(result)$message))
   }
   writeBin(content(result, "raw"), destfile)
--- a/R/bitbucket.R
+++ b/R/bitbucket.R
@@ -3,11 +3,17 @@ isBitbucketURL <- function(url) {
 }
 
 canUseBitbucketDownloader <- function() {
-  all(packageVersionInstalled(devtools = "1.9.1", httr = "1.0.0"))
+  (all(packageVersionInstalled(httr = "1.0.0")) &
+     !is.null(bitbucket_user(quiet = TRUE)) &
+     !is.null(bitbucket_pwd(quiet = TRUE)))
 }
 
 bitbucketDownload <- function(url, destfile, ...) {
-  onError(1, bitbucketDownloadImpl(url, destfile, ...))
+  tryCatch(
+    bitbucketDownloadImpl(url, destfile, ...),
+    error = function(e) {
+      stop(sprintf("Bitbucket request failed with %s", e), call. = FALSE)
+    })
 }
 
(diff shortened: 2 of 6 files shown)


NEWS.md
```diff
@@ -1,5 +1,14 @@
 # Packrat 0.8.0 (UNRELEASED)
 
+- Stop using `devtools::github_pat` to obtain an (optional) `GITHUB_PAT`
+  environment variable value. The `github_pat` function was removed in
+  devtools-2.4.3. (#651)
```
kevinushey, line 5:
```suggestion
  `devtools 2.4.3`. (#651)
```

R/bitbucket.R
```diff
@@ -3,11 +3,17 @@ isBitbucketURL <- function(url) {
...
+     !is.null(bitbucket_user(quiet = TRUE)) &
+     !is.null(bitbucket_pwd(quiet = TRUE)))
 }
 
 bitbucketDownload <- function(url, destfile, ...) {
-  onError(1, bitbucketDownloadImpl(url, destfile, ...))
+  tryCatch(
+    bitbucketDownloadImpl(url, destfile, ...),
+    error = function(e) {
+      stop(sprintf("Bitbucket request failed with %s", e), call. = FALSE)
```
kevinushey, line 15:
```suggestion
      stop(sprintf("Bitbucket request failed with %s", conditionMessage(e)), call. = FALSE)
```

R/bitbucket.R
```diff
@@ -24,15 +30,18 @@ bitbucketDownloadImpl <- function(url, destfile, ...) {
...
   request <- GET(url, auth)
-  if (request$status == 401) {
-    warning("Failed to download package from Bitbucket: not authorized. ",
-            "Did you set BITBUCKET_USERNAME and BITBUCKET_PASSWORD env vars?",
-            call. = FALSE)
-    return(1)
+  if (result$status != 200) {
+    stop(
+      sprintf(
+        "Unable to download package from Bitbucket; check the BITBICKET_USERNAME and BITBUCKET_PASSWORD environment variables: %s",
```
kevinushey, line 36:
```suggestion
        "Unable to download package from Bitbucket; check the BITBUCKET_USERNAME and BITBUCKET_PASSWORD environment variables: %s",
```

R/github.R
```diff
@@ -3,23 +3,51 @@ isGitHubURL <- function(url) {
...
 canUseGitHubDownloader <- function() {
-  all(packageVersionInstalled(devtools = "1.9.1", httr = "1.0.0"))
+  all(packageVersionInstalled(httr = "1.0.0")) & (!is.null(github_pat()))
+}
+
+github_pat <- function(quiet = TRUE) {
+  pat <- Sys.getenv("GITHUB_PAT")
+  if (nzchar(pat)) {
+    if (!quiet) {
+      message("Using GitHub username from envvar GITHUB_PAT")
```
kevinushey, line 13:
```suggestion
      message("Using GitHub PAT from envvar GITHUB_PAT")
```

R/github.R
```diff
@@ -3,23 +3,51 @@ isGitHubURL <- function(url) {
...
 githubDownload <- function(url, destfile, ...) {
-  onError(1, {
-    github_pat      <- yoink("devtools", "github_pat")
-    authenticate    <- yoink("httr", "authenticate")
-    GET             <- yoink("httr", "GET")
-    content         <- yoink("httr", "content")
+  tryCatch(
+    githubDownloadImpl(url, destfile, ...),
+    error = function(e) {
+      stop(sprintf("GitHub request failed with %s", e), call. = FALSE)
```
kevinushey, line 24:
```suggestion
      stop(sprintf("GitHub request failed with %s", conditionMessage(e)), call. = FALSE)
```

R/gitlab.R
```diff
@@ -3,11 +3,17 @@ isGitlabURL <- function(url) {
...
+     !is.null(gitlab_user(quiet = TRUE)) &
+     !is.null(gitlab_pwd(quiet = TRUE)))
 }
 
 gitlabDownload <- function(url, destfile, ...) {
-  onError(1, gitlabDownloadImpl(url, destfile, ...))
+  tryCatch(
+    gitlabDownloadImpl(url, destfile, ...),
+    error = function(e) {
+      stop(sprintf("GitLab request failed with %s", e), call. = FALSE)
```
kevinushey, line 15:
```suggestion
      stop(sprintf("GitLab request failed with %s", conditionMessage(e)), call. = FALSE)
```

R/bitbucket.R
```diff
@@ -3,11 +3,17 @@ isBitbucketURL <- function(url) {
 }
 
 canUseBitbucketDownloader <- function() {
-  all(packageVersionInstalled(devtools = "1.9.1", httr = "1.0.0"))
+  (all(packageVersionInstalled(httr = "1.0.0")) &
```
kevinushey, line 6: nit: use `&&` here if these are scalar logicals

kevinushey (commented):
One issue: if I understand correctly, this PR changes `githubDownload()` and friends to return `TRUE` on success, whereas the previous implementations used something like a status code (`0` for success, `1` for failure).

I think this is the right change but you might need to ensure that this is handled correctly in `downloadWithRetries()` as well.

R/bitbucket.R
kevinushey: I guess this depends on whether you want the error prefix here; e.g.

```
> err <- tryCatch(stop("oh no", call. = FALSE), error = identity)
> sprintf("%s", err)
[1] "Error: oh no\n"
> sprintf("%s", conditionMessage(err))
[1] "oh no"
```

so the existing formulation is probably fine.

aronatkins: Because these eventually get wrapped again, it's probably better to use `conditionMessage`.

aronatkins: After some more testing, I've decided to leave it as-is because we get more context for things like the original `devtools::github_pat` error:

```
Reason: Error: GitHub request failed: Error in get(name, envir = asNamespace(pkg), inherits = FALSE): object 'github_pat' not found
```

aronatkins: I've slightly tweaked the message, but am not using `conditionMessage`.

R/github.R
aronatkins: As above: message tweaked, but not using `conditionMessage`.

R/gitlab.R
aronatkins: As above: message tweaked, but not using `conditionMessage`.

aronatkins:
@kevinushey - `downloadWithRetries` no longer calls the `githubDownload` family of functions. They were always being called outside `downloadWithRetries` whenever the `canUseGitHubDownloader` family of functions was truthy. 

In other words: The code was live and we checked the logic, but it would never actually dispatch to the `githubDownload` function because `getSourceForPkgRecord` always bypasses `downloadWithRetries` when `canUseGitHubDownloader`.

Related: I think we will eventually want `downloadWithRetries` to expose more information about its download errors, but did not want to tackle that with this work.

aronatkins:
Received a verbal +1 from @kevinushey.

PR #653, opened 2021-12-21
cache RemoteType: url similarly to CRAN

### LINKED ISSUE #652: RemoteType: url
Older versions of remotes (prior to 2.0.2) added `RemoteType: url` to package DESCRIPTIONs.

https://github.com/r-lib/remotes/commit/5a92ecab33974b3d70c5947db0bd62fde81dce6d

Packrat should treat this similarly to CRAN packages when determining which fields should be included in the hash computation.

https://github.com/rstudio/packrat/blob/63a95e754d59386f26f5bcad8e7129c9eec145dc/R/cache.R#L38-L46

aronatkins:
Fixes #652

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,9 @@
 
 - Provide more information about authenticated download failures.
 
+- Hash `RemoteType: url` packages as though they came from a CRAN-like
+  repository. Improves cache reuse for packages originally installed by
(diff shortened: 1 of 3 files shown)


aronatkins:
Received verbal 👍 from @kevinushey

PR #654, opened 2022-01-31
ignore packages when recursively enumerating dependencies

aronatkins:
previously, we ignored only direct dependencies

Given a Shiny application having:

```r
library(shiny)
library(emo) # hadley/emo

ui <- fluidPage("So much", emo::ji("dancing"))
server <- function(input, output) {}
shinyApp(ui = ui, server = server)
```

The default set of dependencies:

```
> packrat:::appDependencies(".")
 [1] "R6"          "Rcpp"        "assertthat"  "base64enc"   "bslib"       "cachem"      "commonmark" 
 [8] "crayon"      "digest"      "ellipsis"    "emo"         "fastmap"     "fontawesome" "fs"         
[15] "generics"    "glue"        "htmltools"   "httpuv"      "jquerylib"   "jsonlite"    "later"      
[22] "lifecycle"   "lubridate"   "magrittr"    "mime"        "packrat"     "promises"    "purrr"      
[29] "rappdirs"    "rlang"       "sass"        "shiny"       "sourcetools" "stringi"     "stringr"    
[36] "withr"       "xtable"     
```

Prior to this change, with "emo" and "Rcpp" as ignores, only the top-level "emo" is ignored:

```
> packrat::set_opts(ignored.packages = c("emo","Rcpp"), persist = FALSE)
> packrat:::appDependencies(".")
 [1] "R6"          "Rcpp"        "base64enc"   "bslib"       "cachem"      "commonmark"  "crayon"     
 [8] "digest"      "ellipsis"    "fastmap"     "fontawesome" "fs"          "glue"        "htmltools"  
[15] "httpuv"      "jquerylib"   "jsonlite"    "later"       "lifecycle"   "magrittr"    "mime"       
[22] "packrat"     "promises"    "rappdirs"    "rlang"       "sass"        "shiny"       "sourcetools"
[29] "withr"       "xtable"     
```

After this change, both "emo" and "Rcpp" are ignored:

```
> packrat::set_opts(ignored.packages = c("emo","Rcpp"), persist = FALSE)
> packrat:::appDependencies(".")
 [1] "R6"          "base64enc"   "bslib"       "cachem"      "commonmark"  "crayon"      "digest"     
 [8] "ellipsis"    "fastmap"     "fontawesome" "fs"          "glue"        "htmltools"   "httpuv"     
[15] "jquerylib"   "jsonlite"    "later"       "lifecycle"   "magrittr"    "mime"        "packrat"    
[...]

PR #656, opened 2022-02-08
option setting does not modify other options

### LINKED ISSUE #655: option setters clear other settings
Using the option setter clears other configured options.

```
packrat::set_opts(ignored.packages = c("emo"), persist = FALSE)
packrat::opts$snapshot.recommended.packages(TRUE, persist = FALSE)
packrat::get_opts(c("ignored.packages"))
# => NULL
```

This is using the setter defined by:

https://github.com/rstudio/packrat/blob/c648ce2fb915289d3de5c58eae9b15202a839163/R/options.R#L173-L179

the setter is connected to the `packrat::opts` here:

https://github.com/rstudio/packrat/blob/c648ce2fb915289d3de5c58eae9b15202a839163/R/options.R#L221-L223

aronatkins:
fixes #655

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,9 @@
   repository. Improves cache reuse for packages originally installed by
   `remotes<2.0.2`. (#652)
 
+- The `packrat::opts` single-option setter no longer overwrites previously
+  written state. (#655)
+
 # Packrat 0.7.0
 
 - Fixed an issue where Packrat could inadvertently execute non-R code chunks
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,8 +13,8 @@
   repository. Improves cache reuse for packages originally installed by
   `remotes<2.0.2`. (#652)
 
-- The `packrat::opts` single-option setter no longer overwrites previously
-  written state. (#655)
+- The `packrat::set_opts` function and `packrat::opts` single-option setter no
+  longer overwrite previously written in-memory state. (#655)
 
 # Packrat 0.7.0
 
--- a/R/options.R
+++ b/R/options.R
@@ -176,9 +176,7 @@ make_setter <- function(name) {
     if (missing(x)) {
       return(get_opts(name))
     } else {
-      o <- get_opts()
-      o[[name]] <- x
-      setOptions(o, persist = persist)
+      return(setOptions(setNames(list(x), name), persist = persist))
     }
   }
 }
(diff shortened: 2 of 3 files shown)


kevinushey (approved):
LGTM! Thanks for adding a test as well.

toph-allen:
#### What I did

I smoke tested this and #654 in a merged branch.

I tested the changes by:
- initializing Packrat in a test repo (using `external.packages = "packrat"`, which I assume is a requirement for working with local branches);
- running `packrat:::appDependencies(".")` and `reconnect::writeManifest()` before and after running `packrat::set_opts(ignored.packages = c("emo", "Rcpp", "stringr"))`;
- running a few other packrat functions.

Specific to this change, I also ran the snippet of code from Aron's issue:

```r
packrat::set_opts(ignored.packages = c("emo"), persist = FALSE)
packrat::opts$snapshot.recommended.packages(TRUE, persist = FALSE)
packrat::get_opts(c("ignored.packages"))
```

#### Results

The code snippet from Aron's issue performed as expected.

`reconnect::writeManifest()` correctly ignored packages. Previously, because of option-modification that package was performing, Packrat settings were lost, so the correct behavior indicates the change is working correctly.

I noticed that when the `ignored.packages` option was absent (i.e. `packrat.set_opts(ignored.packages = "")`), all the entries in the manifest had the following fields:

```
      "Repository": "https://cran.rstudio.com/",
      "GithubRepo": null,
      "GithubUsername": null,
      "GithubRef": null,
      "GithubSha1": null,
```

With `packrat.set_opts(ignored.packages = c("emo", "Rcpp", "stringr")`, those fields were missing.

I performed the same check with a build of Packrat only from this branch, not from the combined branch, and the same change occurred. I'm not sure if this is expected, or problematic, so I figured I'd raise it here. The manifests are attached to this comment (with their file extension changed to `.txt`, because GitHub doesn't allow `.json` attachments, I guess).

[manifest-655-ignore.txt](https://github.com/rstudio/packrat/files/8043118/manifest-655-ignore.txt)
[manifest-655-no-ignore.txt](https://github.com/rstudio/packrat/files/8043119/manifest-655-no-ignore.txt)

aronatkins:
The GitHub fields are probably a result of using a data frame to record all the dependencies; once one record requires a field, they all receive that value. 

@toph-allen - Could you file an rsconnect issue to track the presence of those null-value fiends? They're harmless, but maybe worth trimming.

toph-allen:
That makes total sense as to the cause.

I'll add it to my list of issues to file :)

PR #672, opened 2022-06-22
Fix Bitbucket and GitLab download calls in `getSourceForPkgRecord()`

### LINKED ISSUE #671: Bitbucket download flow seems to be broken in `getSourceForPkgRecord()`
In the Bitbucket download flow in `getSourceForPkgRecord()`, execution stops with a `FAILED` message when `status` is `TRUE`. However, this status is returned after a _successful_ download.

https://github.com/rstudio/packrat/blob/1e13f7b582c80a00d701a36303ee0e20196436ca/R/restore.R#L367-L379

The fix should be a single character — prepending `!` to `status` on that line.

toph-allen:
## Intent

Fix archive downloads for Bitbucket and GitLab source archives when they're called in `getSourceForPkgRecord()`.

Fixes #671

## Approach

At first, I just flipped the polarity of the conditional evaluating the `bitbucketDownload()` and `gitlabDownload()` functions' output. But then I saw that the GitHub download workflow had been updated, so I modified Bitbucket and GitLab's download flow to follow GitHub's.

## QA Notes

Still thinking about this.

## Checklist

- [x] Added NEWS entry

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # Packrat 0.9.0 (UNRELEASED)
 
+- Fixed a bug preventing successful downloads of private GitLab and Bitbucket
+  archives during restore. (#671)
 
 # Packrat 0.8.0
 
--- a/R/restore.R
+++ b/R/restore.R
@@ -364,19 +364,19 @@ getSourceForPkgRecord <- function(pkgRecord,
         unlink(srczip, recursive = TRUE)
     }, add = TRUE)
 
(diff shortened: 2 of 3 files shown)


toph-allen:
Note: Local tests seem to not be very broken. We'll see how CI fares.

aronatkins (approved):

toph-allen:
GitLab downloads are still broken. Tracked in #673.

toph-allen:
Other than that, this has been verified.

PR #674, opened 2022-06-24
Support using `GITLAB_PAT` for authentication

### LINKED ISSUE #673: Authentication for private GitLab repos need to use PAT
I'm pretty sure that GitLab's API does not allow authenticating with usernames and password (see [here](https://docs.gitlab.com/ee/api/#authentication)). This prevents Packrat's GitLab downloader from working.

There may be some circumstance or version I'm unaware of (an old self-hosted instance maybe?) where password auth is allowed. But we should probably switch to preferring a `GITLAB_PAT` variable if it exists in the environment.

toph-allen:
## Intent

Fix source downloads from private GitLab repositories by preferentially using `GITLAB_PAT` for authentication.

Fixes #673

## Approach

The GitLab downloader now checks for a `GITLAB_PAT`. If that exists, instead of creating the username/password auth, it adds `Private-Token` to the request's headers.

`GITLAB_USERNAME` and `GITLAB_PASSWORD` were removed. See [this comment](https://github.com/rstudio/packrat/pull/674#issuecomment-1169329634) for why.

## Automated Tests

There were no automated tests for GitLab download functionality before. There is one test that is skipped unless run manually.

This PR adds no tests.

## QA Notes

I validated this against a private repo on my GitLab account. Validation is tricky, requiring a Packrat lock file that restores from a GitLab repository.

To validate:

1. Download and extract this archive: [packrat-with-private-package.zip](https://github.com/rstudio/packrat/files/9003675/packrat-with-private-package.zip)
2. Get access from me to the private package hosted on GitLab that this package attempts to restore.
3. Create a [GitLab PAT](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) and make it available in your terminal's environment as `GITLAB_PAT`.
4. Install this branch's version of Packrat.
5. Start an R session in the archive's directory that contains a `packrat` subdirectory. Run `packrat::restore()`. The environment should restore correctly.

We should also confirm:

- If the `GITLAB_PAT` environment variable is missing, an error message is displayed.
- If `GITLAB_USERNAME` and `GITLAB_PASSWORD` variables are present and `GITLAB_PAT` is absent, the same error message about `GITLAB_PAT` is displayed.

## Checklist

- [x] Added NEWS entry


aronatkins (approved):

toph-allen:
I haven’t found any historical reference to GitLab ever allowing basic password auth in its API, even going back into old issues and StackOverflow posts. I wondered if I was missing some search term. The original implementation was tested against a public repo and followed the Bitbucket implementation (Bitbucket _does_ use username and password).

I think it's likely that GitLab has never supported basic auth with username and password, and that this feature has just never been exercised properly due to `renv`'s availability and Connect not supporting restoring packages from private sources.

@aronatkins , do you think it makes sense to remove the GITLAB_USERNAME and GITLAB_PASSWORD “capability” entirely from packrat? I think that leaving it in risks sending users down a dead end.

toph-allen:
@aronatkins I removed `GITLAB_USERNAME` and `GITLAB_PASSWORD` support in the latest commit.

- After an exhaustive search of the web (StackOverflow, GitLab's own issue tracker and community forum, relevant issues on Packrat and Connect repos, and the RStudio community forum), I found no evidence that GitLab ever supported authenticating calls to its API with username and password, or that anyone had tried to use Packrat in this way. Even five years ago, token-based auth was all I found.
- This was added three years ago, and the work seems to only have been verified with public packages. At that point, `renv` already existed and supported GitLab.
- Because this was mainly only used in Connect, and there were other known barriers to restoring private packages on Connect, this was never discovered.

Removing it from Packrat will avoid sending users down a dead end.

aronatkins (approved):

R/gitlab.R
```diff
@@ -44,39 +44,20 @@ gitlabDownloadImpl <- function(url, destfile, ...) {
...
-#' Looks in env var \code{GITLAB_PASSWORD}
+#' A GitLab PAT
+#' Looks in env var \code{GITLAB_PAT}
 #'
 #' @keywords internal
 #'
-gitlab_pwd <- function(quiet = FALSE) {
-  pwd <- Sys.getenv("GITLAB_PASSWORD")
-  if (nzchar(pwd)) {
+gitlab_pat <- function(quiet = FALSE) {
```
aronatkins, line 54: I just realized that the bitbucket and gitlab accessor functions are exported by packrat. I don't think that it was our intent to allow `packrat::gitlab_pat`. That said, it's out in the wild. :(

CC @kevinushey

toph-allen: I don't think they are actually exported! They don't have the `@export` keyword, they don't appear in `NAMESPACE`, and you need to use `:::` to access them.

toph-allen: They do have documentation generated, but it isn't included in the index, because they include `@keyword internal` in a `roxygen2` comment.

toph-allen:
CI is failing on `oldrel-1` — it looks like network errors in `apt-get` or something like that.

PR #677, opened 2022-07-11
Allow using downloading privately-hosted packages with vendored `renv`

toph-allen:
### Intent

- Allow Packrat to leverage functions in embedded `renv` code to download packages hosted in private repos on GitHub, GitLab, and Bitbucket.
- Where possible, simplify and rationalize the package restore workflow to make future maintenance easier.

### Approach

At a high level, reworked the overall workflow for Git providers in the `getSourceForPkgRecord` function. This involves change to downloader and provider-specific functions.

#### File-by-file notes

Because this involves a lot of code moving around, I took file-by-file notes on changes. These are ordered logically, from highest to lowest-level changes, with links to the files in GitHub's "Files changed" view. Hopefully this makes it easier to navigate.

- [restore.R](https://github.com/rstudio/packrat/pull/677/files#diff-e6bbe2eb45292e12f20b55ee8969ee26a9178e96c8bec58ba9ec941f59e8386e)
   - Many changes in `getSourceForPkgRecord`
      - Change the call to `downloadWithRetries` to account for the error being generated within the function.
      - Within each provider workflow, code has changed significantly.
         - The generation of the archive URL from the package record has been moved to a `[provider]ArchiveUrl` function located in each provider's `.R` file.
         - The logic to decide which downloader to dispatch has been moved to the `[provider]Download` function located in each provider's `.R` file. If an error is returned from the download function, a generic message about the failure is added to the error text here.
         - The `local` block that appended the remote info to each package's `DESCRIPTION` file has been moved to a `modifyRemoteInfo` function.
      - A new `modifyRemoteInfo` function houses code that used to exist in each provider workflow.
         - This may be a bad name, as it is not modifying remote info, but rather adding it to `DESCRIPTION`.
         - The function takes a `src` and `dest` argument, as well as a `remote_info` data frame.
         - The function decompresses the archive at `src`, appends the `remote_info` to the `DESCRIPTION` file, and recompresses the package at `dest`. This is thus used to move the package to its final location.
         - GitHub remote info has been modified to now include a `RemoteType` field. It used not to include this. `renv` adds this field in migration from Packrat, and it exists for Bitbucket and GitLab. Within `modifyRemoteInfo` we use this to determine the field locating the package in a subdirectory of the `tar` archive. If adding this field is bad, we could feasibly identify it another way, e.g. see if any of the field names match `GitHubSubdir`. (In fact, writing this, I think this may be the best way to do this.)
         - This function manages the creation and deletion of `dest` `.tar.gz`.
- [github.R](https://github.com/rstudio/packrat/pull/677/files#diff-eec6078c09526fabb06b30250529d1197262fed97d49d89dbf8e4b2954f979eb), [gitlab.R](https://github.com/rstudio/packrat/pull/677/files#diff-d560bb7554dc48fd7cebfcf448acc5653d96b3bb366845ecc4461122bbfe398c), [bitbucket.R](https://github.com/rstudio/packrat/pull/677/files#diff-7e5f841a9d5ec3e5ac2970d0121faed165eb59f97da6f576d7e8b609e80d151c) (described generically as [provider].R)
   - The top level `[provider]Download` functions now contain the logic for selecting what download function to use. The download function logic has been refactored. If the PAT is present in the environment and an error is encountered, the error message is edited to include a directive to check the appropriate environment variable.
   - `[provider]DownloadImpl` has been renamed to `[provider]DownloadHttr` and now just prints the `httr` status message, as the rest of the error messages are provided elsewhere.
   - A new `[provider]ArchiveUrl` function contains the code code that used to live in `getSourceForPkgRecord`.
- [downloader.R](https://github.com/rstudio/packrat/pull/677/files#diff-1446744a7f6227a98ae750eb4fbf879b31c78f89716981bd2ad5a0990c59437e)
   - Removed code to do with using `libcurl` as a download method, including altering `secureDownloadMethod` and removing `canUseLibCurlDownloadMethod`.
   - Adjusted `downloadWithRetries` so that it stops with an error if it fails, instead of simply not returning TRUE.
      - The exception to this is `renvDownload`. `renv` has internal error handling code which raises errors.
   - Added `renvDownload`.
      - Sets the option `renv.http.useragent` to `"curl"`. Sets environment `RENV_DOWNLOAD_METHOD` to the `method` passed into the function.
      - Calls `renv$download`, returning TRUE if successful.
- [utils.R](https://github.com/rstudio/packrat/pull/677/files#diff-60bef5af43a1bd541299ce9bceeb226afb23deaa5e637b145c85e05ffd552b0c)
   - Added `canUseHttr` function, which really just wraps a check to `packageVersionInstalled` but makes other code more readable, maybe?

### Automated Tests

Multiple automated tests have been added for lower-level functions

- We ensure that `renvDownload` passes its arguments through to `renv$download`. This is partially so that we can directly mock `renvDownload` in other tests to work around a bug in `mockery` that prevents namespaced mocks of nested functions.
- Tests ensure that the `[provider]ArchiveUrl` functions are correct. More importantly, they also ensure that `[provider]Download` passes values through to `renvDownload` correctly when the correct options and environment variables are in play.
- `modifyRemoteInfo` is tested.

### QA Notes

It's difficult to publicly and reproducibly verify end-to-end downloads of private packages. We'll perform verification offline.

### Checklist

Note: Some of this is still TODO

- [x] Update `NEWS`


toph-allen:
Tests started failing with [this commit](https://github.com/rstudio/packrat/pull/677/commits/9bae3a4bbb4a0ce67a64901f106c8cd459b83f0d). I have no idea why the tests that are failing started to fail here, because the failures seem unrelated.

R/downloader.R
```diff
@@ -198,7 +191,36 @@ downloadWithRetries <- function(url, ..., maxTries = 5L) {
...
+
+canUseRenvDownload <- function() {
+  getOption("packrat.authenticated.downloads.use.renv", FALSE) &&
+    identical(secureDownloadMethod(), "curl")
+}
+
+renvDownload <- function(url, destfile, method = inferAppropriateDownloadMethod(url), type = NULL, ...) {
+  debug <- getOption("packrat.renv.debug", FALSE)
+  if (debug) {
+    options(renv.download.trace = TRUE) # TODO remove
```
aronatkins, line 208: When does this TODO happen?

R/downloader.R
```diff
@@ -198,7 +191,36 @@ downloadWithRetries <- function(url, ..., maxTries = 5L) {
...
+  }
+
+  # We temporarily set our user agent to "curl" so that Bitbucket will treat us
+  # like a command line and not a browser. Otherwise, if we make unauthorized
+  # requests to Bitbucket .tar.gz URLs, we get redirects instead of a 401.
+  renv_useragent_option <- options("renv.http.useragent" = "curl")
```
aronatkins, line 214: Should we only modify the user-agent when performing a BitBucket download with renv?

R/downloader.R
```diff
@@ -198,7 +191,36 @@ downloadWithRetries <- function(url, ..., maxTries = 5L) {
...
+  on.exit(options(renv_useragent_option), add = TRUE)
+
+  result <- with_envvar(
+    c(RENV_DOWNLOAD_METHOD = method),
+    renv$download(url = url, destfile = destfile, type = type)
+  )
+  if (identical(result, destfile)) {
```
aronatkins, line 221: Is this return value meaningful? What should happen when `result` and `destfile` are not identical? Explain the contract of this function.

Maybe:
```r
return(identical(result, destfile))
```

aronatkins (approved):
Could you explain the contract of each of the download functions? In particular: Where do they write, who handles deletion of files on error, and how does the caller gauge success (return value versus error, etc).

R/downloader.R
toph-allen: It's happening right after I type this.

toph-allen:
@aronatkins Exhaustive description of download functions in use. Let me know if this checks out.

### Packrat download function contracts

> Could you explain the contract of each of the download functions? In particular: Where do they write, who handles deletion of files on error, and how does the caller gauge success (return value versus error, etc).

In general, I've moved away from having success encoded in return values, and towards leaning entirely on R errors (i.e. `stop()`). Occasionally, callers will intercept this to add additional context.

#### `getSourceForPkgRecord`

- The big restore function. Calls individual provider-specific download methods.
- Creates the *path* for the temporary destination file, named `srczip` at this level. It doesn't create the file — the download implementations do that — but it manages its lifecycle, deleting it on exit if it exists.

#### `githubDownload(url = archiveUrl, destfile = srczip)`, `gitlabDownload(url = archiveUrl, destfile = srczip)`, `bitbucketDownload(url = archiveUrl, destfile = srczip)`

I'm going sometimes refer to all of these collectively as `providerDownload`.

- Top-level provider-specific download function. Responsible for dispatching different download implementations.
- Called by `getSourceForPkgRecord`.
- Takes `url` (the place to retrieve the source from) and `destfile` (the location to place the retrieved source) to an inner download implementation, depending on various environmental factors.
- The file that download functions write to — an arg named named `destfile`, called `srczip` in the caller — is a temporary file. The caller, `getSourceForPkgRecord`, handles its creation and deletion. It is always deleted when `getSourceForPkgRecord` exits. (The downloaded source is moved elsewhere after its `DESCRIPTION` has the remote info appended.)
- Returns nothing if successful. Stops with an error if the download fails.
- When downloading with authentication, it calls the download implementations in a `tryCatch` block. This is so that it can add a note to the error message advising the user to "Check the GITHUB_PAT environment variable." This message is added whatever the underlying error. I kept it because this is what Packrat did before, but I don't necessarily think this is the best thing, as it kinda sounds like we've identified the error and are instructing the user on how to solve it.
- Calls `renvDownload`, `providerDownloadHttr`, or `downloadWithRetries`. It does not check the output of these functions; it assumes that they will raise an error if they fail.

#### `renvDownload(url, destfile, type)`

- Download implementation wrapping `renv$download`
- Called by `providerDownload`.
- `url` and `destfile` correspond to `archiveUrl` and `srczip` respectively. These are their `renv` names.
- `type` is a lowercase string `"github"`, `"gitlab"`, or `"bitbucket"`. `renv` uses this to determine what type of authentication to use. This is hard-coded, created at the moment that `providerDownload` calls `renvDownload`.
- Returns `TRUE` if `renv$download` did not cause an error.
    - At Aron's suggestion, this was computed by `return (identical(result, destfile))`, but looking at the function now, this doesn't actually seem to be semantically meaningful.
    - If we actually care that `result` and `destfile` are identical, it should check them. But I don't think that they *can* diverge. I'm not sure why `renv` has these functions return the `destfile` path. The only time that it's possible for `destfile` to change is replacing backslashes with forward slashes on Windows, which is expected. So I'm going to remove the check.
    Now it just returns `TRUE` to fit in with Packrat conventions, but this value is not checked at runtime. Expects that `renv` will raise an error if something goes wrong.
- Writes to `destfile`, which is managed by the top-level function. `renv` writes to another temporary file internally in some instances, but we rely on it to clean those up.
- I added another check to it. Now, even if `renv$download` finishes without raising an error, if `destfile` does not exist at return time, it will raise the same error that `providerDownloadHttr` functions raise in the same condition.

#### `githubDownloadHttr(url, destfile, ...)`, `gitlabDownloadHttr(url, destfile, ...)`, `bitbucketDownloadHttr(url, destfile, ...)`

Reffering to these singularly/collectively as `providerDownloadHttr`.

- The original secure download function. Called by `providerDownload` if its requirements are met and `renvDownload`'s are not.
- Returns `TRUE` if it succeeds, but also calls `stop()` if any errors are encountered.
- Does not manage any files. The file it writes to, `destfile`, is owned by `getSourceForPkgRecord`.

#### `downloadWithRetries(url, ..., maxTries = 5L)`

I didn't touch much of this, except to add an error at the end of the download function wasn't successful.

- Generic download function, used if we have no provider-specific authentication or cannot call those functions for other reasons.
- Wraps the inner `download` function in a retry pattern. Will raise errors that come from that function, and will also raise an error if the `success` marker is not `TRUE` after `maxTries` attempts.
- Returns `TRUE` if its success marker is `TRUE`.
- Passes `destfile` down to `download` via `...`.

#### `download(url, destfile, method = inferAppropriateDownloadMethod(url)`

Down into old packrat code I didn't touch.

- This is old Packrat code that I haven't touched this cycle.
- Does not manage `destfile`. But it does download to a temporary file that it manages, and that is `unlink`ed on exit.
- Returns `status` if `status`. This needs to be handled by its caller. (`downloadWithRetries` handles it by treating anything except `200` as a failure. But really, this shouldn't be necessary — read on. I think this was to handle an old Windows edge case?)
- Otherwise, if `file.rename(tempfile, destfile)` `0` if everything was successful or `1`.
- Calls a `downloadImpl` method which calls a `downloadFile` method, which wraps `download.file`. `download.file` returns `0` if successful, otherwise a status code, which is how `download` ever gets to renaming its `tempfile`. Down here, success is `0` and failure is `1`, which is flipped elsewhere up the stack. This is another reason that, at the level I worked at, I wanted to rely on errors rather than exit values.

R/downloader.R
toph-allen: Update: I didn't use your suggestion — instead I used:

```r
  if (!file.exists(destfile)) {
    stop("No data received.", call. = FALSE)
  }
  return(TRUE)
```

Because `result` will always be identical to `destfile`, whereas if `renv`'s download method fails `destfile` will not exist.

toph-allen:
@aronatkins I found a bug in the logic in the `githubDownload` etc. functions.

Specifically, the conditional logic is specified such that if GitHub is authenticated but neither `renvDownload` nor `githubDownloadHttr` are available, `downloadWithRetries` is not called as a fallback.

I will submit a fix shortly.

toph-allen:
@aronatkins Changes are in [these commits](https://github.com/rstudio/packrat/pull/677/files/9af579874bd1f397f2e4d8da81efefe0cbde4284..3e09a22c47540c71cb682a03513ee96497f32500).

I explored using the `download() { downloadImpl() }` pattern, but found that it would still require a fair amount of boilerplate in the top-level download function, because there is variation on two axes: download method and provider. The `downloadRenv()` function would have required different logic from `githubDownloadHttr()`, for example, and passing that data around would have made each function call different in the top-level dispatch functions.

What I wound up doing was creating an error handling closure that returns a function based on type (`"github"`, `"gitlab"`, or `"bitbucket"`) to generate error text. It makes the download function logic much clearer than before, with richer error messages that give explicit information about the expected behavior.

I also expanded tests for each provider to cover the expected scenarios, and added tests for the new error message function.

R/bitbucket.R
```diff
@@ -1,22 +1,33 @@
...
-
-canUseBitbucketDownloader <- function() {
-  (all(packageVersionInstalled(httr = "1.0.0")) &&
-     !is.null(bitbucket_user(quiet = TRUE)) &&
-     !is.null(bitbucket_pwd(quiet = TRUE)))
-}
-
+# - Equivalent to other git provider download functions.
+# - Called by `getSourceForPkgRecord` (which manages the lifecycle of
+#   `destfile`).Responsible for dispatching different download implementations
```
aronatkins, line 3: nit: Space after period when starting a sentence.

R/downloader.R
```diff
@@ -244,28 +282,85 @@ secureDownloadMethod <- function() {
...
+      if (bitbucketAuthenticated()) {
+        token_msg <- "BITBUCKET_USERNAME and BITBUCKET_PASSWORD found; check that they are correct."
+      } else {
+        token_msg <- "BITBUCKET_USERNAME and BITBUCKET_PASSWORD environment variables not found."
+      }
+    }
+    advice <- c(advice, token_msg)
+
+    # Info on configuration
+    if (canUseRenvDownload()) {
```
aronatkins, line 342: This control logic feels repetitive; when generating this advice, we know if authentication is in use and what provider was used. Those can all be inputs to this function.

```r
authDownloadAdvice <- function(type, authenticated, downloader) { ... }
```

```r
tryCatch(renvDownload(url, destfile, type = "github"), error = authDownloadAdvice("github", TRUE, "renv")
```

If we pass information about what was used, this function is still useful even if we change the conditions we use to decide which technique is in play.

R/restore.R
```diff
@@ -223,302 +226,149 @@ getSourceForPkgRecord <- function(pkgRecord,
...
-    }
+    archiveUrl <- githubArchiveUrl(pkgRecord)
 
     srczip <- tempfile(fileext = '.tar.gz')
-    on.exit({
-      if (file.exists(srczip))
-        unlink(srczip, recursive = TRUE)
-    }, add = TRUE)
+    on.exit(
+      if (file.exists(srczip)) unlink(srczip, recursive = TRUE),
```
aronatkins, line 233: Is the existence check needed? Per help for `unlink`:

> Not deleting a non-existent file is not a failure, nor is being unable to delete a directory if recursive = FALSE.

R/restore.R
```diff
@@ -223,302 +226,149 @@ getSourceForPkgRecord <- function(pkgRecord,
...
-
-      if (length(pkgRecord$gh_subdir))
-        basedir <- file.path(basedir, pkgRecord$gh_subdir)
-
-      if (!file.exists(file.path(basedir, 'DESCRIPTION'))) {
-        stop('No DESCRIPTION file was found in the archive for ', pkgRecord$name)
-      }
-
-      ghdata <- c(
+    # TODO: The archive url workflow for GH suggests that these fields are
```
aronatkins, line 245: Packrat restore may see lockfiles produced by a variety of different client versions. Package descriptions have had inconsistent, varying GitHub repository annotations over time. The other providers arrived later and have more consistency.

aronatkins (approved):

R/restore.R
toph-allen: You're right. I just brought this over from the old code without questioning whether it was necessary. I'll remove it.

R/restore.R
toph-allen: Yeah, that's what I figured. Removing the comment.

R/downloader.R
toph-allen: This makes a lot of sense, and… it *may* be possible to do it and keep the logic of the download function selection readable.

toph-allen: Done.

toph-allen:
@aronatkins, @ChaitaC and I just validated this PR together on 2022-10-17.

Issue #648, opened 2021-12-02
Internal R tar is not handling long filenames.


galachad:
Hi pacrat is using `tar="internal"`. It's perfect solution for compatibility between different systems but internal R tar is not handling long directory names.
https://github.com/rstudio/packrat/blob/3df1d62f77e98f846b11636b4a3203b925ec1028/R/restore.R#L481

Impact: RStudio products, Windows users

Example: Installing package from gitlab with long slug is not working. (https://gitlab.com/adam.forys/long-test-slug-skeleton)

Why it's happening? 

Gitlab endpoint `{gitlab_host}/api/v4/projects/{repository}/repository/archive?sha={sha}`
generate `tar.gz` file that contains directory created as `repository-slug-sha-sha/` so it gave `long-test-slug-skeleton-958296dbbbf7f1d82f7f5dd1b121c7558604809f-958296dbbbf7f1d82f7f5dd1b121c7558604809f` and `untar(tar="internal")` returns error `Error in file(name, "wb") : cannot open the connection`.

Installing package by packrat for repository slug that has more that 16 characters will fail.

Because packrat is soft deprecated it's only impact RStudio Connect/shinyapps etc. (Rstudio product). 

This problem is not an issue for `renv` as `renv` is detecting system tar. 

My recommendation is replace `"internal"` with `Sys.getenv("TAR", "internal")`.
```
untar(srczip, exdir = scratchDir, tar = Sys.getenv("TAR", "internal"))
```
After this change untar will be using system tar when set. 

Question: 
Is any chance this change will be apply in next Rstudio Connect release?

aronatkins:
Here is an example `packrat.lock`:

```
PackratFormat: 1.4
PackratVersion: 0.8.1
RVersion: 3.6.3
Repos: CRAN=https://cran.rstudio.com/

Package: skeleton
Source: gitlab
Version: 1.0.1
Hash: 612f8f6d5a930a63ffe057599cad2343
RemoteRepo: long-test-slug-skeleton
RemoteUsername: adam.forys
RemoteRef: HEAD
RemoteSha: 958296dbbbf7f1d82f7f5dd1b121c7558604809f
RemoteHost: gitlab.com
```

When restored:

```r
packrat::restore()
# => Installing packrat (0.8.1) ... 
# => 	OK (built source)
# => Installing skeleton (1.0.1) ... 
# => Error in file(name, "wb") : cannot open the connection
```

The full trace:

```
[...]

PR #678, opened 2022-07-20
Use internal tar in untar only when TAR is not set

galachad:
The TAR variables is set by R process if tar binary is available. I can imagine that in some system TAR will not be discover property but then standard `internal` will be used.  

It's better to use system tar that internal R tar in case where tar files are not compatible with internal R tar (#648).

This pr is changing only untar param to default behavior where `tar = Sys.getenv("TAR", "internal")`.

Reported long time ago in #648


toph-allen:
This is resolved in #679

PR #679, opened 2022-07-22
Update how `tar` binaries are discovered

### LINKED ISSUE #648: Internal R tar is not handling long filenames.

toph-allen:
This PR incorporates #678, but goes further, so will supersede that PR.

Fixes #648 

### Intent

Packrat has previously forced the use of R's internal tar implementation. This is buggy in a number of situations, including any repository with a slug > 16 characters.

Note: The code is ready for review, but QA notes are incomplete.

### Approach

Influenced by the approach taken in `renv`, we now define a function `tar_binary()` to encapsulate slightly more complex logic:

- If `TAR` is set, respect that;
- If on Unix, look for `tar` on the `PATH`;
- If on Windows, look for the system `tar` and use that if it exists (see [`renv` implementation](https://github.com/rstudio/renv/blob/2bea5cef7cb6c57e43580cf0a182cbe59ef3536a/R/tar.R#L14-L21))
- Otherwise, return `"internal"` and maybe emit a warning about long filenames.

I put the function in `env.R`, because it gathers information about the environment.

There were a number of locations where `tar()` or `untar()` were called with `internal`, and this PR changes all of them.

### Automated Tests

This PR adds automated tests. Note that it adds a dependency on `mockery` for the tests; I'm not sure if we're ok with this.

There are five test functions that test the logical flow of the function in the following cases:

- A `TAR` variable is defined returns the contents of that variable
- We're on Unix and we can find a `tar` binary on the `PATH` returns that binary
- We're on Unix and we can't find a `tar` binary on the `PATH` returns `"internal"`
- We're on Windows and we find system `tar` returns the system `tar`
- We're on Windows and we can't find system `tar` returns `"internal"`

### QA Notes

You can validate that `packrat::restore()` still works by installing the the version of the package from this branch and running `packrat::restore()` in an R session in the top level directory extracted from this .zip file: [simple-rmd.zip](https://github.com/rstudio/packrat/files/9253334/simple-rmd.zip)

This will perform a Packrat restore. I'm not sure if there's a way to validate that the download path has been exercised from `restore()` though.

You can also run `packrat:::tar_binary()` and see that it prints the expected path for your environment. For me, this looks like:

```r
> packrat:::tar_binary()
[1] "/usr/bin/tar"
```

### DIFF
--- a/R/env.R
+++ b/R/env.R
@@ -1,4 +1,36 @@
-# Tools for storing state in environment variables.
+# Tools for getting info about the execution environment
+
+tar_binary() <- function() {
+  # If TAR is specified in the environment, use that.
+  tar <- Sys.getenv("TAR", unset = NA)
+  if (!is.na(tar)) {
+    return(tar)
+  }
+
+  # If we're on Unix, look for a tar binary on the PATH.
+  if (is.unix()) {
+    return(Sys.which("tar"))
+  }
+
+  # If we're on Windows, look for the system tar binary.
+  if (is.windows()) {
+    root <- Sys.getenv("SystemRoot", unset = NA)
+    if (is.na(root)) {
+      root <- "C:/Windows"
+    }
+    tar <- file.path(root, "System32/tar.exe")
+    if (file.exists(tar)) {
+      return(tar)
+    }
+  }
+
+  # Return internal only as a fallback with a warning.
+  warning("No external tar binary found. Using R's internal TAR, which may cause failures with long filenames.")
+  return("internal")
+}
+
+
+# Tools for storing state in environment variables. Possibly unused.
 getenv <- function(x) {
   strsplit(Sys.getenv(x, unset = ""), .Platform$path.sep, fixed = TRUE)[[1]]
 }
--- a/R/env.R
+++ b/R/env.R
@@ -1,6 +1,6 @@
 # Tools for getting info about the execution environment
 
-tar_binary() <- function() {
+tar_binary <- function() {
   # If TAR is specified in the environment, use that.
   tar <- Sys.getenv("TAR", unset = NA)
   if (!is.na(tar)) {
--- a/R/env.R
+++ b/R/env.R
@@ -9,7 +9,10 @@ tar_binary <- function() {
 
   # If we're on Unix, look for a tar binary on the PATH.
   if (is.unix()) {
-    return(Sys.which("tar"))
+    tar <- file.path(Sys.which("tar"))
+    if (file.exists(tar)) {
+      return(tar)
+    }
   }
 
   # If we're on Windows, look for the system tar binary.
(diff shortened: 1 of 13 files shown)


R/env.R
```diff
@@ -1,4 +1,38 @@
...
+      root <- "C:/Windows"
+    }
+    tar <- file.path(root, "System32/tar.exe")
+    if (file.exists(tar)) {
+      return(tar)
+    }
+  }
+
+  # Return internal only as a fallback with a warning.
+  warning("No external tar binary found. Using R's internal TAR, which may cause failures with long filenames.")
```
aronatkins, line 31: How frequently will this warning be emitted when using the `"internal"` tar?

aronatkins (approved):

aronatkins (commented):
This needs NEWS.

ChaitaC:
Packrat
```
RemoteType: github
RemoteHost: api.github.com
RemoteRepo: packrat
RemoteUsername: rstudio
RemoteRef: toph-untar-handling
RemoteSha: 108183392a8d733ed6bf66da65afab5b6de1c11c
GithubRepo: packrat
GithubUsername: rstudio
GithubRef: toph-untar-handling
GithubSHA1: 108183392a8d733ed6bf66da65afab5b6de1c11c
NeedsCompilation: no
Packaged: 2022-08-09 18:31:24 UTC; chaitamacpro
Built: R 4.1.1; ; 2022-08-09 18:31:25 UTC; unix

ChaitaC:
verified on packrat branch`toph-untar-handling`

PR #682, opened 2022-08-04
Add environment masking to Packrat

toph-allen:
### Intent

Mask certain environment variables from subprocesses launched by Packrat.

### Approach

There are two ways an environment variable can be added to the list of environment variables to mask:
- A predefined list of variables is built-in. This can be disabled by setting the option `packrat.mask.git.service.envvars` to `FALSE`.
- Users can define a character vector of additional variable names to mask in the option `packrat.masked.envvars`.

These are compiled in a function called `envvar_mask()`, which returns a named character vector of NAs with the names of variables to mask. This is used in the `R()` function that launches R subprocesses.

Other changes:
- A change was made to `system_check()` to add the argument `return_output = FALSE`, which captures and returns the output of the command, rather than returning `TRUE` upon success. This is used to facilitate tests that launch subprocesses and make assertions about the output.
- Renamed `is.named()`, which checks that all elements of a vector are named, to `all.named()`.

#### Open question

- Maybe the envvar mask should be loaded in `system_check()`. This is lower-level and can launch any process, whereas `R()` is for launching R processes. In practice, they are completely coupled.

### Automated Tests

Tests work one level out from the `envvar_mask()` function, launching an R process to run `R -e 'Sys.getenv()'` to examine the resulting environment. There are three tests:
- Test that the default list of variables is masked with the option unset
- Test that the default list of variables is ignored when the appropriate option is set
- Test that other variables can be masked

I wrote these tests before I moved the code to its own `envvar_mask()` function. It would be possible to test that inner code, but this outer test will fail if the inner code does not function.

### QA Notes

Install the test version of the package.

We're going to examine our environment with `Sys.getenv()`, and then run an R subprocess via Packrat and look at the environment seen by that subprocess. This will confirm that the new variable masking options are working correctly.

#### 1. Git service account variables are masked by default, but can be unmasked with an option.

In an R session, set an environment variable named `GITHUB_PAT`. The second line here should print out the result.

```r
Sys.setenv(GITHUB_PAT = "very_secret")
Sys.getenv("GITHUB_PAT")
```
```
[1] "very_secret"
```

Next, we'll see if it appears in a subprocess. This will print out a fair amount of text, including the R startup text. The result here is truncated. The important part is that the variable `very_secret` should not appear below the line where our function call is printed out.

```r
packrat:::R("-e 'Sys.getenv(\"GITHUB_PAT\")'", return_output = TRUE)
```
```
[20] "> Sys.getenv(\"GITHUB_PAT\")"                                 
[21] "[1] \"\""                                                     
```

Next, we'll turn off Git variable masking and see if the variable does appear.

```r
options(packrat.mask.git.service.envvars = FALSE)
packrat:::R("-e 'Sys.getenv(\"GITHUB_PAT\")'", return_output = TRUE)
```
```
[20] "> Sys.getenv(\"GITHUB_PAT\")"                                 
[21] "[1] \"very_secret\""                                          
```

#### 2. Arbitrary environment variables we provide can be masked with an option.

Set two variables and examine them locally.

```r
Sys.setenv(PLANT_BASED_PAT = "veggie_patty")
Sys.setenv(MEAT_EATERS_OPTION = "beef_patty")
Sys.getenv(c("PLANT_BASED_PAT", "MEAT_EATERS_OPTION"))
```
```
   PLANT_BASED_PAT MEAT_EATERS_OPTION 
    "veggie_patty"       "beef_patty" 
```

Check them in a subprocess.

```r
packrat:::R("-e 'Sys.getenv(c(\"PLANT_BASED_PAT\", \"MEAT_EATERS_OPTION\"))'", return_output = TRUE)
```
```
[20] "> Sys.getenv(c(\"PLANT_BASED_PAT\", \"MEAT_EATERS_OPTION\"))" 
[21] "   PLANT_BASED_PAT MEAT_EATERS_OPTION "                       
[22] "    \"veggie_patty\"       \"beef_patty\" "                   
```

Set new Packrat option and check that they are absent from the subprocess output.

```r
options(packrat.masked.envvars = c("PLANT_BASED_PAT", "MEAT_EATERS_OPTION"))
packrat:::R("-e 'Sys.getenv(c(\"PLANT_BASED_PAT\", \"MEAT_EATERS_OPTION\"))'", return_output = TRUE)
```
```
[20] "> Sys.getenv(c(\"PLANT_BASED_PAT\", \"MEAT_EATERS_OPTION\"))" 
[21] "   PLANT_BASED_PAT MEAT_EATERS_OPTION "                       
[22] "                \"\"                 \"\" "                   
```

#### 3. Verify that these options work together

Turn Git variable masking back on. Change the user mask option to only specify one of the variables. Confirm that all our variables still exist.

```r
options(packrat.mask.git.service.envvars = NULL)
options(packrat.masked.envvars = "PLANT_BASED_PAT")
Sys.getenv(c("GITHUB_PAT", "PLANT_BASED_PAT", "MEAT_EATERS_OPTION"))
```
```
        GITHUB_PAT    PLANT_BASED_PAT MEAT_EATERS_OPTION 
     "very_secret"     "veggie_patty"       "beef_patty" 
```

Check these variables in a subprocess. Only `MEAT_EATERS_OPTION` should be there.

```r
packrat:::R("-e 'Sys.getenv(c(\"GITHUB_PAT\", \"PLANT_BASED_PAT\", \"MEAT_EATERS_OPTION\"))'", return_output = TRUE)
```
```
[20] "> Sys.getenv(c(\"GITHUB_PAT\", \"PLANT_BASED_PAT\", \"MEAT_EATERS_OPTION\"))"
[21] "        GITHUB_PAT    PLANT_BASED_PAT MEAT_EATERS_OPTION "                   
[22] "                \"\"                 \"\"       \"beef_patty\" "             
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # Packrat 0.9.0 (UNRELEASED)
 
+- Packrat now masks environment variables commonly used for Git service account
+  authentication from subprocesses used to install packages. This behavior can
+  be disabled by setting the option `packrat.unmask.git.service.vars` to `TRUE`.
+- Users can now mask arbitrary environment variables from the subprocess that
+  run package installation tasks, by setting the option `packrat.masked.envvars`
+  to a character vector of variable names to mask.
 
 # Packrat 0.8.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,10 +2,10 @@
 
 - Packrat now masks environment variables commonly used for Git service account
   authentication from subprocesses used to install packages. This behavior can
-  be disabled by setting the option `packrat.unmask.git.service.vars` to `TRUE`.
-- Users can now mask arbitrary environment variables from the subprocess that
+  be disabled by setting the option `packrat.mask.git.service.envvars` to `FALSE`.
+- Users can mask additional arbitrary environment variables from the subprocess that
   run package installation tasks, by setting the option `packrat.masked.envvars`
-  to a character vector of variable names to mask.
+  to a character vector of variable names to mask. These variables are masked whether or not `packrat.mask.git.service.envvars` has been set to `FALSE`.
 
 # Packrat 0.8.1
 
--- a/R/install.R
+++ b/R/install.R
@@ -126,35 +126,12 @@ R <- function(options, path = tempdir(), env_vars = NULL, ...) {
     on.exit(set_path(old))
   }
 
-  # Mask Git service tokens unless told not to.
-  git_token_vars <- if (!getOption("packrat.unmask.git.service.vars", FALSE)) {
-    c(
-      "GITHUB_PAT" = NA,
-      "GITLAB_PAT" = NA,
-      "BITBUCKET_USERNAME" = NA,
-      "BITBUCKET_USER" = NA,
-      "BITBUCKET_PASSWORD" = NA,
-      "BITBUCKET_PASS" = NA,
-      # Varnames that may have been used previously
-      "GITHUB_USERNAME" = NA,
-      "GITHUB_USER" = NA,
-      "GITHUB_PASSWORD" = NA,
-      "GITHUB_PASS" = NA,
-      "GITLAB_USERNAME" = NA,
-      "GITLAB_USER" = NA,
-      "GITLAB_PASSWORD" = NA,
-      "GITLAB_PASS" = NA
-    )
-  } else {
-    NULL
-  }
-
   in_dir(
     path,
     system_check(
       cmd = r_path,
       args = options,
-      env = c(r_env_vars(), env_vars, git_token_vars),
+      env = c(r_env_vars(), env_vars, envvar_mask()),
       ...
     )
   )
@@ -172,6 +149,40 @@ r_env_vars <- function() {
     "NOT_CRAN" = "true",
     "TAR" = auto_tar()
   )
+}
+
+envvar_mask <- function() {
+  # Mask tokens unless told not to.
+  git_token_vars <- if (!getOption("packrat.unmask.git.service.vars", FALSE)) {
+    c(
+      "GITHUB_PAT",
+      "GITLAB_PAT",
+      "BITBUCKET_USERNAME",
+      "BITBUCKET_USER",
+      "BITBUCKET_PASSWORD",
+      "BITBUCKET_PASS",
+      # Varnames that may have been used previously
+      "GITHUB_USERNAME",
+      "GITHUB_USER",
+      "GITHUB_PASSWORD",
+      "GITHUB_PASS",
+      "GITLAB_USERNAME",
+      "GITLAB_USER",
+      "GITLAB_PASSWORD",
+      "GITLAB_PASS"
+    )
+  } else {
+    NULL
+  }
+
+  user_specified_vars <- getOption("packrat.masked.envvars", NULL)
+
+  all_vars <- c(git_token_vars, user_specified_vars)
+
+  envvar_mask <- as.character(rep(NA, length(all_vars)))
+  names(envvar_mask) <- all_vars
+
+  return(envvar_mask)
 }
 
 auto_tar <- function() {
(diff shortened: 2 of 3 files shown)


NEWS.md
```diff
@@ -1,5 +1,11 @@
 # Packrat 0.9.0 (UNRELEASED)
 
+- Packrat now masks environment variables commonly used for Git service account
+  authentication from subprocesses used to install packages. This behavior can
+  be disabled by setting the option `packrat.unmask.git.service.vars` to `TRUE`.
```
aronatkins, line 5: What about using `packrat.mask.git.service.vars` with a default of `TRUE` and using `FALSE` to stop variable masking. This option feels more like "disable the feature" rather than "enable the absence of the feature".

NEWS.md
```diff
@@ -1,5 +1,11 @@
...
+- Users can now mask arbitrary environment variables from the subprocess that
+  run package installation tasks, by setting the option `packrat.masked.envvars`
```
aronatkins, line 7: The two settings use different terminology to talk about environment variables. Should we consistently use either `vars` or `envvars`?

Explain how this set of character vectors interacts with the `git.service.vars` option.

R/install.R
```diff
@@ -126,7 +126,15 @@ R <- function(options, path = tempdir(), env_vars = NULL, ...) {
     on.exit(set_path(old))
   }
 
-  in_dir(path, system_check(r_path, options, c(r_env_vars(), env_vars), ...))
+  in_dir(
+    path,
+    system_check(
+      cmd = r_path,
+      args = options,
+      env = c(r_env_vars(), env_vars, envvar_mask()),
```
aronatkins, line 134: No caller passes `env_vars` to `R(...)`. Should we ditch that argument?

R/install.R
```diff
@@ -632,7 +678,7 @@ wrap_command <- function(x) {
 }
 
 set_envvar <- function(envs, action = "replace") {
-  stopifnot(is.named(envs))
+  stopifnot(all.named(envs))
```
aronatkins, line 681: Could you add a comment explaining the behavior of this function (and also to `with_envvar`? The behavior of `UNSET_ME=NA` is a little subtle.

tests/testthat/test-install.R
```diff
@@ -0,0 +1,83 @@
...
+
+test_that("User-specified environment variables are masked from an R subprocess", {
+  new_envvars <- c(
+    "MY_SPECIAL_PAT" = "veggie_patty",
+    "MEAT_EATERS_OPTION" = "beef_patty"
+  )
+  prior_envvars <- set_envvar(new_envvars, "replace")
+  on.exit(set_envvar(new_envvars, "replace"), add = TRUE, after = FALSE)
+
+  # First, we should check that our environment variables appear in the subprocess
```
aronatkins, line 69: Should we extract this as a separate test? It confirms that unmasked variables (either through the option or Git) are available to subprocesses.

Other additional permutation: A subprocess that validates all of the following at once:
* Git variables are masked
* User-enumerated variables are masked
* Other variables are visible
This test would confirm that we don't have a "everything or nothing" masking problem.

aronatkins (commented):
A few minor adjustments to option naming and tests, but otherwise, looks grand.

toph-allen:
Awesome! Thanks for the review; I'll get this set for QA this afternoon.

NEWS.md
toph-allen: Agreed. I went back and forth on which way to go, and with the distance of a weekend, your suggestion seems much clearer.

NEWS.md
toph-allen: Agreed, I should pick one. I think I'll use `envvars` because it's more precise, and there is [precedent](https://stat.ethz.ch/R-manual/R-devel/library/base/html/EnvVar.html).

Each option has an independent effect on the vector of variables to be is masked:
- The `git.service.envvars` governs whether that set of variables are added to the vector.
- And any user-specified variables are added to the vector.

I'll update the NEWS entry to make that clearer.

R/install.R
toph-allen: Yes, I'm all for ditching unnecessary complexity :)

tests/testthat/test-install.R
toph-allen: Yes, I'll work on these test changes after addressing the above comments (so that names etc. are updated).

aronatkins (approved):
The PR description still references the old setting names.

toph-allen:
@aronatkins Tests are also currently failing — I paused working on that last night

ChaitaC:
Packrat 
```
RemoteUsername: rstudio
RemoteRef: toph-env-masking
RemoteSha: b14a93fc0a99dd0e9327851bc493a1b8df8b5ebf
GithubRepo: packrat
GithubUsername: rstudio
GithubRef: toph-env-masking
GithubSHA1: b14a93fc0a99dd0e9327851bc493a1b8df8b5ebf
NeedsCompilation: no
Packaged: 2022-08-09 22:43:02 UTC; chaitamacpro
Built: R 4.1.1; ; 2022-08-09 22:43:03 UTC; unix
```

toph-allen:
Verified on Zoom with Chaita

PR #683, opened 2022-08-09
Update vendored `renv`; update script to allow taking tagged versions

toph-allen:
This PR updates the embedded version of `renv` used by Packrat. Since the current `HEAD` on the `renv` repo is not passing CI, I took the previous version. To do this, I modified the `tools-sync-renv.R` script to allow passing a single argument, like so:

```bash
R -f tools/tools-sync-renv.R --args 0.15.5-48
```

If there are zero arguments, the script will proceed as before. If there is one argument, it uses that in a modified `git checkout` command. If there is more than one argument, it stops and emits a warning.

### QA notes

These are taken from the last "bump renv" PR.

We want to ensure that this change doesn't break dependency detection, which is currently the only place we use `renv`. One way to exercise this is to deploy a simple Shiny app to an RStudio Connect instance.

After installing the version of Packrat in this PR, deploy the following Shiny app to a Connect instance and ensure that it runs.

```R
library(shiny)

ui <- fluidPage(
    titlePanel("Old Faithful Geyser Data"),
    sidebarLayout(
        sidebarPanel(
            sliderInput("bins",
                        "Number of bins:",
                        min = 1,
                        max = 50,
                        value = 30),
            div("username", verbatimTextOutput("user")),
            div("groups", verbatimTextOutput("groups"))
        ),

        mainPanel(
            plotOutput("plot")
        )
    )
)

server <- function(input, output, session) {
  output$plot <- renderPlot({
    x    <- faithful[, 2]
    bins <- seq(min(x), max(x), length.out = input$bins + 1)

    hist(x, breaks = bins, col = 'darkgray', border = 'white')
  })

  output$user <- renderText(session$user)
  output$groups <- renderText(session$groups)
}

```


tools/tools-sync-renv.R
```diff
@@ -1,3 +1,8 @@
+# This script updates the vendored version of renv. It must be run from the
+# Packrat root directory. By default, uses the current version of renv on main.
+# You can pass a single tag as an argument after --args to use that version.
+# R -f /tools/test.R --args 0.15.5-48
```
aronatkins, line 4: Are renv tags ever overwritten? It appears as if @kevinushey generally bumps the version, but I don't know if that's true of every commit.

Should we suggest using commit references rather than tags?

aronatkins (approved):
Idle musing about the renv-updater script.

tools/tools-sync-renv.R
```diff
@@ -7,12 +12,29 @@ desc <- read.dcf("DESCRIPTION", all = TRUE)
 if (!identical(desc$Package, "packrat"))
   stop("script must be run from packrat root directory")
 
+# determine which tag we're downloading, if any
```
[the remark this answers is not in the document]
toph-allen: I don't feel that strongly. I made this choice for two reasons:

1. It looks like it's [more complex](https://stackoverflow.com/a/43136160) to shallow clone a ref than a tag or branch
2. The old TODO commend indicated that the author was thinking about branches ¯\_(ツ)_/¯

Also

3. `renv` has about one tag per commit on `main`, so it's functionally not too different

tools/tools-sync-renv.R
```diff
@@ -7,12 +12,29 @@ desc <- read.dcf("DESCRIPTION", all = TRUE)
...
 
-# check out the latest renv sources
-# TODO: clone a specific version / tag
-system("git clone --depth 1 https://github.com/rstudio/renv")
+if (is.null(tag)) {
+  # check out the latest renv sources
+  system("git clone --depth 1 https://github.com/rstudio/renv")
+} else if (length(tag) == 1) {
+  # check out the tagged version
+  git_clone_cmd <- paste0("git clone --branch ", tag, " --depth 1 https://github.com/rstudio/renv")
```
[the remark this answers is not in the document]
toph-allen: Good point!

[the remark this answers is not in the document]
toph-allen: Good point! I'll include the argument that was used to clone a specific tag if one was provided

tools/tools-sync-renv.R
toph-allen: I'll see what it would be like to use commit references.

toph-allen: To do this would require changing the conditional flow to the following, and would only accept *full shas*, not abbreviations.

```r
if (is.null(ref)) {
  # check out the latest renv sources
  system("git clone --depth 1 https://github.com/rstudio/renv")
  twd <- setwd("renv")
} else if (length(ref) == 1) {
  # check out the tagged version
  dir.create("renv")
  twd <- setwd("renv")
  system("git init")
  system("git remote add origin https://github.com/rstudio/renv")
  system(paste("git fetch --depth 1 origin", ref))
  system("git checkout FETCH_HEAD")
}

gitCommitHash <- system("git rev-parse HEAD", intern = TRUE)
setwd(twd)
```

I'm going to leave this here for now, and not make this change unless you feel strongly that this is better. I feel comfortable treating tags on `renv` as reasonable ways to find commits, and as you wrote in another comment, the header comment in the resulting `renv.R` file mentions the commit that was used, so… I don't _think_ this causes too much of a problem.

toph-allen:
@aronatkins I updated logic to remove one of the conditional statements by just capturing the command to use. Now the message at the top of `renv.R` includes the `git clone` command used and the timestamp at which it was called. The message also includes the `renv` version and the hash.

inst/resources/renv.R
```diff
@@ -1,6 +1,6 @@
 #
-# Auto-generated by 'tools/tools-sync-renv.R'.
-# Cloned from renv using tag: 0.15.5-48.
+# Auto-generated by 'tools/tools-sync-renv.R' at 2022-08-09 16:19:51
```
aronatkins, line 2: I don't feel like the timestamp adds information and could do without it.

aronatkins (approved):
No need to adjust further. We can continue to iterate on the renv sync script as we take additional updates.

ChaitaC:
verified on packrat branch `toph-update-renv`

```
RemoteRef: toph-update-renv
RemoteSha: 7d423359b75a1fd0214ccbfb2078d7a16dee812e
GithubRepo: packrat
GithubUsername: rstudio
GithubRef: toph-update-renv
GithubSHA1: 7d423359b75a1fd0214ccbfb2078d7a16dee812e
NeedsCompilation: no
Packaged: 2022-08-10 14:37:32 UTC; chaitamacpro
Built: R 4.1.1; ; 2022-08-10 14:37:33 UTC; unix
```

Screen Shot 2022-08-10 at 9 38 27 AM

PR #685, opened 2022-08-15
build ignore expressions as expected by renv rather than rooted paths

### LINKED ISSUE #684: packrat (renv) dependency lookup inadvertently ignores projects in directories prefixed with `/data/`.
We have a situation where our system temp directory is actually a symlink to `/data/tmp`, and the renv default exclusion rules cause all inferred packages to be ignored when trying to deploy content to RStudio Connect.

While in our case `/data` is the first directory after the root, it seems this will happen if a directory is called `data` anywhere in the tree. I think ideally these exclusions would match only child directories of the _project_ root. This should be a relatively minimal example.

``` r
dir.create("data/foo", recursive = TRUE)
writeLines("library(glue)", "data/foo/test.R")
options(renv.renvignore.exclude = packrat:::ignoresForRenv(packrat:::opts$ignored.directories()))
path <- normalizePath("data/foo")
packrat:::renv$dependencies(path, root = path)
#> Finding R package dependencies ... Done!
#> [1] Source  Package Require Version Dev    
#> <0 rows> (or 0-length row.names)
```

<sup>Created on 2022-08-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

cc @aronatkins, who I believe added this code in https://github.com/rstudio/packrat/pull/647

A workaround for us is to set `options(packrat.dependency.discovery.renv = FALSE)`, but ideally users wouldn't need to resort to this :)

aronatkins:
fixes #684

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -168,7 +168,7 @@ dirDependenciesRenv <- function(dir) {
     on.exit(Sys.setenv(RENV_PROFILE = profile), add = TRUE)
   }
 
-  absDir <- normalizePath(dir)
+  absDir <- normalizePath(dir, winslash = "/")
 
   old_ignored_packages <- options("renv.settings.ignored.packages" = opts$ignored.packages())
   on.exit(do.call(options, old_ignored_packages), add = TRUE)
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -126,7 +126,9 @@ dirDependencies <- function(dir) {
 # Note: The "/data/" and "/inst/" directories are ignored by default.
 #
 # See: https://github.com/rstudio/renv/pull/866
-ignoresForRenv <- function(ignoredDirectories) {
+#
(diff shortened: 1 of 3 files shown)


aronatkins:
@jimhester Would you be able to test this change in your environment?

jimhester:
@aronatkins I can confirm this PR fixes the issue in our environment. Thanks for working on it!

toph-allen (approved):

R/dependencies.R
```diff
@@ -163,16 +168,17 @@ dirDependenciesRenv <- function(dir) {
     on.exit(Sys.setenv(RENV_PROFILE = profile), add = TRUE)
   }
 
+  absDir <- normalizePath(dir)
```
kevinushey, line 171: I think you'll want `winslash = "/"` here, since `renv` always tries to parse paths using forward slashes. (Or, the path / slash normalization should happen in `ignoresForRenv`.)

PR #688, opened 2022-11-07
Document TAR handling

toph-allen:
### Intent

Better document TAR binary discovery heuristic.

### Approach

Added a new paragraph to the `?restore` docs describing the heuristic. Link to `?restore` from the `?bundle` documentation. From a scan of the code, these were the two documented functions that wind up tarring or untarring things — some but not all of which happened automatically when calling `devtools::document()`.

Some reflowing of roxygen comments.

### DIFF
--- a/R/packrat.R
+++ b/R/packrat.R
@@ -14,7 +14,8 @@
 #' packages your project depends on.
 #' \item \strong{Reproducible}: Packrat records the exact package versions you
 #' depend on, and ensures those exact versions are the ones that get installed
-#' wherever you go.}
+#' wherever you go.
+#' }
 #'
 #' Use \code{\link{init}} to create a new packrat project,
 #' \code{\link{snapshot}} to record changes to your project's library, and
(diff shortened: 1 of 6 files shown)


R/packrat.R
```diff
@@ -269,59 +270,77 @@ initImpl <- function(project = getwd(),
...
+#'   \code{restore} works only on the private package library created by
+#'   packrat; if you have other libraries on your path, they will be unaffected.
+#'
+#'   The \code{restart} parameter will only result in a restart of R when the R
+#'   environment packrat is running within makes available a restart function
+#'   via \code{getOption("restart")}.
+#'
+#'   To install packages hosted in private repositories on GitHub, GitLab, and
+#'   Bitbucket, you must either set the option
+#'   \code{packrat.authenticated.downloads.use.renv} to \code{TRUE} and ensure
```
aronatkins, line 305: This change isn't included in the NEWS; was that because we were keeping it as an internal option until it got some mileage? I would be happy to see it flip to enabled by default and leave the option in case folks want to revert to the `httr` behavior.

For comparison, we do not explain that `packrat.dependency.discovery.renv` can be used to revert to the old dependency scanning workflow; it remains (temporarily) in case folks find some issue with the new code.

aronatkins (approved):
OK with these changes, but also open to making `packrat.authenticated.downloads.use.renv` the default (with accompanying NEWS entry).

R/packrat.R
toph-allen: As far as I recall, in our discussions, we had considered enabling it by default, but decided to have it default to `FALSE` to change as little as possible.

This preference is already mentioned in the unreleased news.

toph-allen: If this option defaulted to `TRUE`, `renv` code would only be used to download when Git provider authentication environment variables are found and when `curl` is available. That will include _all_ downloads from supported Git providers, not just ones in private repos.

I would be fine enabling it, but I feel like we already made this decision, and I can't recall the reasoning.

toph-allen: This conversation was resolved in our Sync meeting today. We're going to enable this feature in a separate PR in the future.

PR #695, opened 2023-01-04
Enable Windows CI on an old release

toph-allen:


toph-allen:
Closing in favor of #697

PR #697, opened 2023-01-04
enable windows ci

aronatkins:


aronatkins (commented):
@toph-allen  and I paired through these changes / CI failure fixes. The Windows CI step was added after CRAN-submission Windows tests failed.

PR #698, opened 2023-01-05
initial lintr configuration

aronatkins:
runs lintr during CI.

most existing lintr warnings are suppressed rather than reacting to them.


toph-allen (approved):

aronatkins:
force-pushed after update with `main` to obtain Windows CI and related fixes.

PR #703, opened 2023-02-17
Fix issue adjusting `DESCRIPTION` files for SCM-backed packages in subdirectories of remote repos

toph-allen:
### Intent

Fix an issue where Packrat would fail to restore packages from subdirectories of repositories.

### Approach

There were a few complex nested calls to `as.data.frame()`, `c()`, and `list()` spread across `getSourceForPkgRecord()` that I attempted to simplify in an earlier update. I simplified them incorrectly, causing `subdir` properties to be dropped.

Because we don't have end-to-end tests set up, and this transformation was part of a long function, no tests caught the issue. I moved the transformations into their own function and added unit tests.

Stepping back, this problem stems from the overall complexity of Packrat's workflows. This solution does not directly address the complexity. It adds a bit of indirection to make the outer-level workflow clearer, but it doesn't reduce the overall complexity of the data flows in the restore process.

### QA Notes

End-to-end validation should involve restoring a Packrat environment that contains a package in a subdirectory. I have already tested that myself, and I'll walk through it with the package's maintainers before this merges.

### DIFF
--- a/R/remote-info.R
+++ b/R/remote-info.R
@@ -1,32 +1,34 @@
 getRemoteInfo <- function(pkgRecord) {
   if (pkgRecord$source == "github") {
-    remote_info <- as.data.frame(
-      c(
-        list(
-          RemoteType     = pkgRecord$source,
-          GithubRepo     = pkgRecord$gh_repo,
-          GithubUsername = pkgRecord$gh_username,
-          GithubRef      = pkgRecord$gh_ref,
-          GithubSHA1     = pkgRecord$gh_sha1
-        ),
-        c(GithubSubdir   = pkgRecord$gh_subdir)
-      ),
-      stringsAsFactors = FALSE
+    return(
+      as.data.frame(
+        c(
+          list(
+            RemoteType     = pkgRecord$source,
+            GithubRepo     = pkgRecord$gh_repo,
+            GithubUsername = pkgRecord$gh_username,
+            GithubRef      = pkgRecord$gh_ref,
+            GithubSHA1     = pkgRecord$gh_sha1
+          ),
+          c(GithubSubdir   = pkgRecord$gh_subdir)
+        ), stringsAsFactors = FALSE
+      )
     )
   } else {
-    remote_info <- as.data.frame(
-    c(
-      list(
-        RemoteType     = pkgRecord$source,
-        RemoteHost     = pkgRecord$remote_host,
-        RemoteRepo     = pkgRecord$remote_repo,
-        RemoteUsername = pkgRecord$remote_username,
-        RemoteRef      = pkgRecord$remote_ref,
-        RemoteSha      = pkgRecord$remote_sha
-      ),
-      c(RemoteSubdir = pkgRecord$remote_subdir)
-    ),
-    stringsAsFactors = FALSE
+    return(
+      as.data.frame(
+        c(
+          list(
+            RemoteType     = pkgRecord$source,
+            RemoteHost     = pkgRecord$remote_host,
+            RemoteRepo     = pkgRecord$remote_repo,
+            RemoteUsername = pkgRecord$remote_username,
+            RemoteRef      = pkgRecord$remote_ref,
+            RemoteSha      = pkgRecord$remote_sha
+          ),
+          c(RemoteSubdir = pkgRecord$remote_subdir)
+        ), stringsAsFactors = FALSE
+      )
     )
   }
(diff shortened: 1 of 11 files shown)


R/remote-info.R
```diff
@@ -0,0 +1,32 @@
+getRemoteInfo <- function(pkgRecord) {
+  if (pkgRecord$source == "github") {
+    remote_info <- as.data.frame(
```
aronatkins, line 3: The assigned variable is unnecessary; you can directly return the data frame.

R/remote-info.R
```diff
@@ -0,0 +1,32 @@
...
+      c(
+        list(
+          RemoteType     = pkgRecord$source,
+          GithubRepo     = pkgRecord$gh_repo,
+          GithubUsername = pkgRecord$gh_username,
+          GithubRef      = pkgRecord$gh_ref,
+          GithubSHA1     = pkgRecord$gh_sha1
+        ),
+        c(GithubSubdir   = pkgRecord$gh_subdir)
```
aronatkins, line 12: Having the subdir fields outside the main list is subtle; maybe add a comment explaining why this approach is necessary.

aronatkins (approved):

R/remote-info.R
toph-allen: I made this change.

R/remote-info.R
toph-allen: Yes, this is exactly what tripped *me* up last time. And I _knew_ why they were outside the main list. I almost feel like doing this much more verbosely but more explicitly would be better.

R/remote-info.R
```diff
@@ -0,0 +1,32 @@
+getRemoteInfo <- function(pkgRecord) {
+  if (pkgRecord$source == "github") {
+    remote_info <- as.data.frame(
+      c(
+        list(
+          RemoteType     = pkgRecord$source,
```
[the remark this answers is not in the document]
toph-allen: It didn't have them [before](https://github.com/rstudio/packrat/blob/d152f12602e2595e3cfe3c73acc9a1a2de3ecad0/R/restore.R#L307). (Also note that the SHA field is differently shaped.

R/remote-info.R
toph-allen: I tried to explain what's going on.

R/remote-info.R
```diff
@@ -0,0 +1,43 @@
+# This code is difficult to read and has caused problems in the past. Heed my
+# warning. This so that as.data.frame() is given a list. The *_subdir fields are
+# missing in most cases. If they were included in the main list() calls, list()
+# would include a field with that name with a NULL value.
+#
+# Creating a list and then concatenating the possibly-NULL subdir fields means
+# that they are NULL, they will not appear at all in the resulting list at all.
+# The resulting data frame is later appended to the DESCRIPTION file, so this is
+# desirable.
+getRemoteInfo <- function(pkgRecord) {
```
aronatkins, line 10: The old GitHub handling had a very simple (to my eye) shape:

https://github.com/rstudio/packrat/blob/d152f12602e2595e3cfe3c73acc9a1a2de3ecad0/R/restore.R#L307-L315

Using that as inspiration, that could give you:

```r
return(as.data.frame(as.list(c(
             RemoteType     = pkgRecord$source,
             GithubRepo     = pkgRecord$gh_repo,
             GithubUsername = pkgRecord$gh_username,
             GithubRef      = pkgRecord$gh_ref,
             GithubSHA1     = pkgRecord$gh_sha1,
             GithubSubdir = pkgRecord$gh_subdir
)), stringsAsFactors = FALSE)
```

aronatkins (approved):
I'm not convinced that this is the most obvious way to materialize the data.frame, but the tests show that we're getting back the expected value, which is enough to address the problem.

R/remote-info.R
toph-allen: It's almost certainly not the most obvious way to materialize it — this was the original way it was materialized, and that same impulse led me to try to simplify it and break it.

I might have another go at simplifying — and just do it in a more verbose, obvious way — now that we have the tests.

toph-allen: Maybe I didn't try `as.list`, just `list`.

aronatkins (approved):

toph-allen:
Rerunning validation against this new commit.

aronatkins (approved):

PR #726, opened 2024-01-02
installPkg: do no work when using cached package and annotate safely

### LINKED ISSUE #720: packrat restore race condition
The use of packrat in Posit Connect can hit race-conditions during restore. It seems packrat directly writes to/reads from the DESCRIPTION file of packages and if two Connect apps are using the same package and are updated/redeployed at the same time, we could hit a race-condition where one or both updates fail. i.e.

```
2023/08/04 14:09:06.743139312 # Installing required R packages with `packrat::restore()` ---------------------
2023/08/04 14:09:07.220617061 Installing DBI (1.1.1) ... 
2023/08/04 14:09:07.245971037 Error in annotatePkgDesc(pkgRecord, project, lib): nrow(content) == 1 is not TRUE
2023/08/04 14:09:07.245980057 
2023/08/04 14:09:07.246015294 Unable to fully restore the R packages associated with this deployment.
2023/08/04 14:09:07.246019737 Please review the preceding messages to determine which package
2023/08/04 14:09:07.246055930 encountered installation difficulty and the cause of the failure.
2023/08/04 14:09:07.249423406 Warning message:
2023/08/04 14:09:07.249481533 In packrat::restore(overwrite.dirty = TRUE, prompt = FALSE, restart = FALSE) :
2023/08/04 14:09:07.249487850   The most recent snapshot was generated using R version 4.1.1
```
In this case a retry succeeded, so this was not an instance of a corrupt packrat cache.

This is especially fragile in a Connect instance with many git-backed deployments as multiple apps could be redeployed concurrently when Git.Concurrency>1.

aronatkins:
First, the `installPkg()` function no longer does additional work when the package was discovered in the cache. This avoids the most common situation where multiple processes performing package installation can race to the `annotatePkgDesc()` and `moveInstalledPackageToCache()` calls.

Second, `annotatePkgDesc()` writes its update to a temporary file before renaming to the target `DESCRIPTION`. This avoids a (now more rare) situation where a second process might see an incompletely written `DESCRIPTION` file.

As a side-effect, only the originally installing actor will annotate the `DESCRIPTION`. Previously, an update occurred even when the package was used from the cache, which overwrote an existing annotation.

Update: After thinking about this problem a bit more, realized that the root cause to #720 is because `annotatePkgDesc()` was previously writing into the package cache. After calling `restoreWithCopyFromCache()`, the library contains a symlink into the cache. Calling `annotatePkgDesc()` reads through the symlink and modifies the in-cache `DESCRIPTION`. This is why we saw multiple actors reading/writing the same file.

Fixes #720

### DIFF
--- a/R/restore.R
+++ b/R/restore.R
@@ -395,14 +395,6 @@ annotatePkgDesc <- function(pkgRecord, project, lib = libDir(project)) {
   write_dcf(content, descFile)
 }
 
-# Annotate a set of packages by name.
-annotatePkgs <- function(pkgNames, project, lib = libDir(project)) {
-  records <- searchPackages(lockInfo(project), pkgNames)
-  lapply(records, function(record) {
-    annotatePkgDesc(record, project, lib)
-  })
-}
-
 # Takes a vector of package names, and returns a logical vector that indicates
 # whether the package was not installed by packrat.
 installedByPackrat <- function(pkgNames, lib.loc, default = NA) {
--- a/R/restore.R
+++ b/R/restore.R
@@ -574,7 +574,6 @@ installPkg <- function(pkgRecord,
     annotatePkgDesc(pkgRecord, project, lib),
     error = function(e) {
       unlink(pkgInstallPath, recursive = TRUE)
-      return(e)
     }
   )
 
--- a/R/restore.R
+++ b/R/restore.R
@@ -394,7 +394,9 @@ annotatePkgDesc <- function(pkgRecord, project, lib = libDir(project)) {
   # Write it out using a temporary file so DESCRIPTION is never partial.
   tmpf <- tempfile(tmpdir = dirname(descFile))
   write_dcf(content, tmpf)
-  file.rename(tmpf, descFile)
+  if (!file.rename(tmpf, descFile)) {
+    stop("Unable to rename annotated package DESCRIPTION")
+  }
 }
 
 # Takes a vector of package names, and returns a logical vector that indicates
--- a/R/restore.R
+++ b/R/restore.R
@@ -427,7 +427,6 @@ installPkg <- function(pkgRecord,
                        repos,
                        lib = libDir(project))
 {
-  pkgSrc <- NULL
   type <- "built source"
   needsInstall <- TRUE
 
@@ -525,14 +524,11 @@ installPkg <- function(pkgRecord,
     })
   }
 
-  if (is.null(pkgSrc)) {
+  if (needsInstall) {
     # When installing from github/bitbucket/gitlab or an older version, use the cached source
     # tarball or zip created in snapshotSources
     pkgSrc <- file.path(srcDir(project), pkgRecord$name,
                         pkgSrcFilename(pkgRecord))
-  }
-
-  if (needsInstall) {
 
     if (!file.exists(pkgSrc)) {
       # If the source file is missing, try to download it. (Could happen in the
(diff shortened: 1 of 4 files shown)


R/restore.R
```diff
@@ -391,16 +391,10 @@ annotatePkgDesc <- function(pkgRecord, project, lib = libDir(project)) {
...
-# Annotate a set of packages by name.
-annotatePkgs <- function(pkgNames, project, lib = libDir(project)) {
-  records <- searchPackages(lockInfo(project), pkgNames)
-  lapply(records, function(record) {
-    annotatePkgDesc(record, project, lib)
-  })
+  # Write it out using a temporary file so DESCRIPTION is never partial.
+  tmpf <- tempfile(tmpdir = dirname(descFile))
+  write_dcf(content, tmpf)
+  file.rename(tmpf, descFile)
```
kevinushey, line 397: I believe this will emit a warning and return false on failure -- do we need to handle that scenario?

kevinushey (approved):

R/restore.R
aronatkins: You're right. I had dismissed the need to handle the error because the directory is known to be writable and we are not performing a cross-file-system move, but it's probably worth checking the return value.

toph-allen (approved):
No major comments from a code read. Questions that came up were answered elsewhere in the code.

One conceptual question: Do we know why the description was updated when installing from the cache? Are there any times where un-annotated descriptions can enter the cache (e.g. a package was installed via install.packages without Packrat) where we'd want to add the annotation? And if so, would the atomic updates (via renaming a temp file) be sufficient to fix the problem?

kevinushey:
> Are there any times where un-annotated descriptions can enter the cache (e.g. a package was installed via install.packages without Packrat) where we'd want to add the annotation?

A long time ago, Packrat tried to distinguish between packages which were installed by Packrat itself, versus packages that were installed by the user. The intention here was to have a way of differentiating between packages that were "managed" or "known" to Packrat, and use that to customize the UI presented to the user in certain scenarios (e.g. helping them choose whether to snapshot or restore).

In practice, users found the behavior hard to understand, so that was phased out over time. I'm not sure if we still make use of the InstallAgent annotations anywhere in the Packrat sources, but if we do, it'd be nice to drop those once and for all.

R/restore.R
```diff
@@ -598,35 +570,38 @@ installPkg <- function(pkgRecord,
   }
 
   # Annotate DESCRIPTION file so we know we installed it
-  annotatePkgDesc(pkgRecord, project, lib)
+  withCallingHandlers(
+    annotatePkgDesc(pkgRecord, project, lib),
+    error = function(e) {
+      unlink(pkgInstallPath, recursive = TRUE)
+      return(e)
```
kevinushey, line 577: Nit: calling handlers don't alter control flow, so I believe the `return()` usage here is superfluous / ignored.

kevinushey (approved):

R/restore.R
aronatkins: Thanks. Confirmed that this is the case.

aronatkins:
> I'm not sure if we still make use of the InstallAgent annotations anywhere in the Packrat sources

We do still use `InstallAgent`. See calls to `installedByPackrat()`.

aronatkins:
> Are there any times where un-annotated descriptions can enter the cache (e.g. a package was installed via install.packages without Packrat) where we'd want to add the annotation? And if so, would the atomic updates (via renaming a temp file) be sufficient to fix the problem?

Users do not enroll packages in the cache; only Packrat performs cache additions and cache-uses while installing a package. I suppose it's possible that a user could add something to the Packrat library, but that is separate from cache management.

PR #730, opened 2025-01-31
downloads: alternate archive layouts

### LINKED ISSUE #583: Add support for Artifactory archive types
In [RTFACT-17392](https://www.jfrog.com/jira/browse/RTFACT-17392), Artifactory has a bug dealing with how CRAN archives.   While that is indeed a bug, I have found in 3.5.1 install.packages, varying implementation of REMOTES, and/or versions of packrat, each package attempt to solve these varying implementations differently.  Some need to reference a "archive.rds".  Others must have a package.rds while others sufficiently use a PACKAGE file.

The ask and the associated PR is allowing the loosening of these restrictions to be backwards compatible.  Remotes has a similar [bug and PR](https://github.com/r-lib/remotes/issues/440)

aronatkins:
this is not the most efficient approach, but attempts multiple repository archive layouts before giving up.

reduces the number of download attempts before giving up. generally, a single retry is enough. also reduces the cost of additional layout attempts.

fixes #583


kevinushey (approved):
LGTM, although I'd suggest providing some way to opt-out of this behavior (or opt-in to a specific formatter) just in case an escape hatch is needed.

aronatkins:
@kevinushey - Do you think it's worth adapting a memoizing technique like you use in renv? I can certainly add an opt-out that falls back to the CRAN-only approach.

kevinushey:
> Do you think it's worth adapting a memoizing technique like you use in renv? I can certainly add an opt-out that falls back to the CRAN-only approach.

I'm sort of on the fence... how close are we to phasing out internal use of packrat? It might be fine to just take this PR as-is and leave it alone.

toph-allen (approved):
Also on the fence about memoizing vs just using the naive less efficient implementation. I'm going to approve as I think it'd be _fine_ as-is, but if implementing a simple memoize function was easy, maybe that'd be better? Not very helpful, I know.

PR #743, opened 2025-08-21
Reformat with air

karawoo:
This is the result of `air format $(find . -name "*.R")`. Obviously a pretty large diff, but all tests continue to pass.


inst/vendor/renv.R
aronatkins, line 1: This file (`inst/vendor/renv.R`) is code-generated by renv and should not be formatted.

scenarios/git/scenario-git-1.R
aronatkins, line 1: Something to consider separately: The `scenarios/` directory looks like some manual tests, but I can't see evidence that they are used. Could we remove them? CC @kevinushey

aronatkins (approved):

aronatkins:
My approval is conditional on reverting `inst/vendor/renv.R`. Ideally, we would tell `air` to ignore that file.

karawoo:
@aronatkins I've reverted those changes and updated the air.toml to exclude only that file. I actually hadn't noticed the air.toml before, that explains why `air format .` didn't initially work.

aronatkins (approved):
Feel free to (separately) add a workflow to enforce air formatting going forward, like https://github.com/rstudio/rsconnect/blob/main/.github/workflows/format-suggest.yaml

andrew-bierman:
@copilot

PR #746, opened 2025-09-19
Prioritize `RemoteSubdir` over `GithubSubdir` for GitHub remotes

### LINKED ISSUE #740: Consider both `Github*` and `Remote*` package description fields if one is missing
It is possible for packrat to receive a lockfile that has incomplete package description metadata. For example, `GithubSubdir` could be missing while `RemoteSubdir` is present. In this scenario right now, packrat will fail to restore the package because if it's a github package it only looks in the `Github*` fields for the subdirectory information and doesn't find it. We could also look at the `Remote*` fields as a fallback (or vice versa).

It's not totally obvious which should take precedence, but I believe `Remote*` is the newer way of specifying this information, so maybe we should migrate to preferring `Remote*` fields and only using `Github*` if `Remote*` isn't present.

See also https://github.com/rstudio/renv/issues/2165

karawoo:
Currently we only look in the `GithubSubdir` field for subdirectory information for GitHub remotes. `RemoteSubdir` is the more modern place to record the subdirectory. This PR looks in `RemoteSubdir` first, and falls back to `GithubSubdir` if that's `NULL`.

Fixes #740

### DIFF
--- a/R/restore.R
+++ b/R/restore.R
@@ -1152,12 +1152,15 @@ appendRemoteInfoToDescription <- function(src, dest, remote_info) {
     basedir <- scratchDir
   }
 
-  # Determine the true package root
-  if (remote_info$RemoteType == "github") {
-    remote_subdir <- remote_info$GithubSubdir
-  } else {
+  # Determine the true package root. RemoteSubDir is the more modern way of
+  # recording this info, but in some cases GithubSubDir may be used instead (or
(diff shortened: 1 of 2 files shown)


aronatkins (approved):
Do any of the other "subdir" references need to be adjusted?

```bash
git grep -i GithubSubdir
git grep -i gh_subdir
```

andrew-bierman:
@copilot

PR #755, opened 2026-05-29
fix and add integration tests for installing from bitbucket/gitlab/github 

### LINKED ISSUE #754: Add integration tests for remote package restore
## Context

Posit Connect has been removing its integration tests that verify restoring packages from remote sources (posit-dev/connect#39649, posit-dev/connect#39655). Testing remote download behavior is more suited to Packrat itself since it owns that logic.

## Proposal

Add integration tests that verify `packrat::restore()` works with remote sources. Modeled on how renv does it ([test-retrieve.R](https://github.com/rstudio/renv/blob/b7a73826b20f5bbded7f785f0b47b112e4b7b6d6/tests/testthat/test-retrieve.R#L50-L65)):

1. Create a minimal test package hosted under a Posit-controlled account/org
2. Write testthat tests that restore from a lockfile pointing to the remote source
3. Assert the package installs and can be loaded
4. Mark as `skip_on_cran()` / skip when offline

### Open questions

- Should the test package live under `rstudio/` on GitHub? Elsewhere?
- Should these be opt-in (skipped by default) given they require network access?

## References

- Connect PR removing Bitbucket test: posit-dev/connect#39649
- Connect issue removing GitLab test: posit-dev/connect#39655
- renv's approach: https://github.com/rstudio/renv/blob/main/tests/testthat/test-retrieve.R

karawoo:
Closes #754

Fixes the existing bitbucket test which attempted to install falsy from a bitbucket repo that no longer exists. Replaces with the skeleton package that renv also [tests against](https://github.com/rstudio/renv/blob/b7a73826b20f5bbded7f785f0b47b112e4b7b6d6/tests/testthat/test-retrieve.R#L50-L65). 

Adds a test for installing from github, and for all tests adds some additional logic to ensure that the package is actually installed and loadable.

### DIFF
--- a/tests/testthat/projects/falsy-bitbucket/deps.R
+++ b/tests/testthat/projects/falsy-bitbucket/deps.R
@@ -1,1 +1,0 @@
-library(falsy)
--- a/tests/testthat/projects/falsy-github/deps.R
+++ b/tests/testthat/projects/falsy-github/deps.R
@@ -1,0 +1,1 @@
+library(falsy)
--- a/tests/testthat/projects/falsy-github/packrat/packrat.lock
+++ b/tests/testthat/projects/falsy-github/packrat/packrat.lock
@@ -1,0 +1,18 @@
+PackratFormat: 1.4
+PackratVersion: 0.5.0.11
+RVersion: 3.5.1
+Repos: CRAN=https://cran.rstudio.com/
+
+Package: falsy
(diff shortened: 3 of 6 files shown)


aronatkins (approved):
CC @kevinushey - Packrat is testing against your skeleton repos.

PR #756, opened 2026-07-13
fix handling of concurrent package installs

karawoo:
Concurrent R processes installing the same package into a shared package cache can collide in `moveInstalledPackageToCache()`, where the losing process fails with "failed to restore package ... may be lost from cache".

The function already attempted to handle this with a `file.exists()` check before renaming into the cache, but a competing process can populate the cache entry between the check and the rename. This PR moves the check to after the rename fails, where there is no window. 

Also, only roll back a backed-up cache entry when a backup was actually made. The rollback previously ran unconditionally, so a fresh package that failed to insert reported "may be lost from cache" instead of "failed to copy package to cache".

The race window is extremely small, so the tests mock `file.rename` to simulate the losing interleaving.


edavidaja (approved):

R/cache.R
```diff
@@ -251,26 +251,28 @@ moveInstalledPackageToCache <- function(
   tempPath <- tempfile(tmpdir = dirname(cachedPackagePath))
   on.exit(unlink(tempPath, recursive = TRUE), add = TRUE)
   if (all(dir_copy(packagePath, tempPath))) {
-    # check to see if the cached package path exists now; if it does,
```
aronatkins, line 254: The `dir_copy` is a comparatively slow operation and likely to see two processes racing. Unfortunately, because there is time between `file.exists` and `file.rename`, our previous code resolved the most common type of races, but did not eliminate them.

Because `file.rename` is atomic, using it alone is a better guard than our previous `file.exists` before `file.rename`.

👍

R/cache.R
aronatkins, line 245: Note: This earlier rename always fails in Connect deployments because the R library and cache are beneath separate filesystems (mounts).

aronatkins, line 230: Note: This earlier rename is both good (removing a potentially old cache entry so it can be replaced with just-rebuilt package contents) and bad (many processes may see the need to build this package, separately call `moveInstalledPackageToCache` and move aside just-created cache entries).

Everything after the cache checks in `installPkg` leads to these races that we're resolving, because package-building and cache-population are expensive, which makes it more likely that multiple restore operations are doing that work at the same time.

Have you explored writing "burn in" tests of `installPkg` to see what other problems still linger? This early rename concerns me; maybe `overwrite = TRUE` is a mistake?

aronatkins (commented):
Thanks so much for improving this cache population logic; it has been a source of fragility.

R/cache.R
karawoo: Thanks for calling this out. In practice, nothing ever calls `moveInstalledPackageToCache` with `overwrite = TRUE`. So I think we are okay here today, but we could remove the `overwrite` argument entirely and just avoid this whole trap for the future.

karawoo: sorry I mean nothing uses `fatal = TRUE`, so this branch doesn't get reached

aronatkins: Yeah; neither `fatal` nor `overwrite` ever use non-default values, so we can remove both of them (as suggested in #757).

PR #757, opened 2026-07-15
remove overwrite arg from `moveInstalledPackageToCache`

karawoo:
See https://github.com/rstudio/packrat/pull/756#discussion_r3587548514. That comment flagged that `moveInstalledPackageToCache`'s backup/overwrite path (renaming an existing cache entry aside before replacing it) can race with another process populating that same entry. This function is called in two places in `installPkg` and neither passes `fatal = TRUE`. This path was only reachable if `fatal = TRUE` and `overwrite = TRUE` so removing it should be safe avoids a potential trap in the future.

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -257,8 +257,11 @@ moveInstalledPackageToCache <- function(
     return(symlinkPackageToCache(packagePath, cachedPackagePath))
   }
 
-  # rename failed; copy to temporary destination in same directory
-  # and then attempt to rename from there
+  # rename failed. this can happen when the library and cache live on different
+  # filesystems, when a competing process created the cache entry first, or on
+  # other OS errors (permissions, disk full). copy to a temporary destination in
+  # the same directory as the target (so the final rename stays within one
+  # filesystem) and attempt to rename from there
   tempPath <- tempfile(tmpdir = dirname(cachedPackagePath))
   on.exit(unlink(tempPath, recursive = TRUE), add = TRUE)
   if (all(dir_copy(packagePath, tempPath))) {
--- a/R/cache.R
+++ b/R/cache.R
@@ -194,6 +194,27 @@ symlinkPackageToCache <- function(packagePath, cachedPackagePath) {
   return(cachedPackagePath)
 }
 
+# Record that the freshly-built package at packagePath is being discarded in
+# favor of an existing cache entry (either one that predates this call, or one
+# a competing process just won the race to populate). Logs the reason when
+# cache verbosity is on, and if cacheCopyStatus is supplied, sets its $type
+# field so callers can report what actually happened instead of the outcome
+# they expected before caching was attempted.
+discardBuildForCacheEntry <- function(packageName, reason, cacheCopyStatus) {
+  if (isVerboseCache()) {
+    message(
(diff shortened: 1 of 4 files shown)


R/cache.R
```diff
@@ -198,43 +198,27 @@ symlinkPackageToCache <- function(packagePath, cachedPackagePath) {
 # package into the cache and replace the original directory with a symbolic
 # link into the package cache.
 #
-# If the package already exists inside the cache, overwrite=TRUE causes
-# replacement of the cached content while overwrite=FALSE with fatal=FALSE
-# uses the cached package. Using overwrite=TRUE with fatal=TRUE will err.
```
karawoo, line 203: just a note because it threw me off briefly, I believe this comment was inaccurate and the combination that errors is `fatal = TRUE` and `overwrite = FALSE`

aronatkins: Thanks. This comment was either incorrect when originally written or grew stale as the implementation evolved.

Should we remove `fatal` as well?

R/cache.R
```diff
@@ -198,43 +198,27 @@ symlinkPackageToCache <- function(packagePath, cachedPackagePath) {
...
-  if (file.exists(cachedPackagePath)) {
-    if (!file.rename(cachedPackagePath, backupPackagePath)) {
-      stop(
-        "failed to back up package '",
-        packageName,
-        "'; cannot safely copy to cache"
-      )
-    }
-    on.exit(unlink(backupPackagePath, recursive = TRUE), add = TRUE)
+    return(symlinkPackageToCache(packagePath, cachedPackagePath))
```
aronatkins, line 248: There are a number of paths through `moveInstalledPackageToCache` which discard the just-built package. Should we add logging for these situations? The computed `type` in `installPkg` is also misleading.

> built source but i lost the race. discarded the build and symlinked cache.

Too wordy, but saying something more could help subsequent debugging.

R/cache.R
aronatkins: Also: `overwrite` was a lie. Maybe always?

aronatkins (approved):

R/cache.R
karawoo: sure I'll remove `fatal` too

R/cache.R
karawoo: I updated to take the `cacheCopyStatus` argument that other related functions use, so that we can report the right info back to `installPkg`. Added some messaging as well.

R/cache.R
aronatkins: Is it worth enumerating the reasons why the rename might fail? Other than OS errors (full disk, no permissions):
1. library and cache are on different file systems.
2. cached location exists (created by another)

aronatkins (approved):

karawoo:
lintr released today and is objecting to `<<-`, going to just bypass with `# nolint` since it's unrelated to this PR.

aronatkins:
> lintr released today and is objecting to `<<-`, going to just bypass with `# nolint` since it's unrelated to this PR.

😭 Thanks. They haven't published updated release notes, yet. https://github.com/r-lib/lintr/blob/main/NEWS.md#lintr-340

